PR #6, opened 2022-10-28
RFC: A new schedule way for distributed lookup join

chenzl25:


liurenjie1024:
I have an optimization to this, for distributed lookup join, we can add an exchange to the build side(outer side):
```
LookupJoin
   HashExchange(key = index)
       IndexTable
```

And the lookup join needs to buffer all data from index table. This is quite similar to dynamic filter approach, and there are several advantages:
1. We can avoid sending scan tasks to all inner table when build side is not large.
2. Compare to current approach, we also avoid sending duplicate task same vnode in primary table.

liurenjie1024:
Per offline discussion, we agree that we can use this proposal for now.

liurenjie1024 (approved):

skyzh:
I feel like this is the same idea as my proposed streaming delta join. We should push data instead of doing remote lookup. If we can make this possible, it will probably also be easier for us to re-enable streaming delta join (as now we have upstream hash shard information in optimizer).

skyzh:
... and the overall design looks great to me!

PR #7, opened 2022-10-28
2-phase approx count distinct agg

soundOfDestiny:


TennyZhuang:
Generally LGTM

rfcs/0007-two-phase-approx-count-distinct.md
```diff
@@ -0,0 +1,36 @@
...
+
+2-phase Approx Count Distinct Agg which uses the bucket_id as distribution key to calculate vnode for 2-phase agg, as jon-chuang proposed in https://github.com/risingwavelabs/risingwave/issues/3414.
+
+## Motivation
+
+Currently the `partial_to_total_agg` of `Approx Count Distinct Agg` is `Sum0`, but it is a bad estimation because $Cardinality(S1 \cup S2) \neq Cardinality(S1) + Cardinality(S2)$.
+
+## Design
+
+* Alter group-by key of input into `bucket_id || group_by_keys`. (For example, in GlobalSimpleAgg whose `group_by_keys` is `None`, the new group-by key will be `bucket_id`.) We need to leverage `StreamExchange` to alter distribution key accordingly if necessary.
```
st1page, line 20: will we introduce a new expression to evaluate the bucket_id? It sounds a little strange to me because no other things need that expression

rfcs/0007-two-phase-approx-count-distinct.md
```diff
@@ -0,0 +1,36 @@
...
+* The partial agg is hash agg.
+* The partial agg should maintain bucket state in HyperLogLog algorithm. The implementation of the partial agg is detailedly described in https://github.com/risingwavelabs/risingwave/issues/3414.
```
st1page, line 22: sounds the design still needs to store lots of KV pairs with all the buckets. :thinking:

soundOfDestiny: > sounds the design still needs to store lots of KV pairs with all the buckets. 🤔

If we don't shuffle, even in append-only case the change row will be large. Let's postpone 2-phase if it's not acceptable.

soundOfDestiny: > sounds the design still needs to store lots of KV pairs with all the buckets. 🤔

yes but We only need to store once. In original design, we will use `group_by_count` times as much storage as the design here.

xxchan (commented):
What's the status of this RFC?

soundOfDestiny:
> What's the status of this RFC?

Let's close it for now because the bucket state is too large.

PR #2, opened 2022-10-26
RFC: The WatermarkFilter and StreamSort operator

TennyZhuang:
Signed-off-by: TennyZhuang <zty0826@gmail.com>

PR #9, opened 2022-10-31
add watermark_strategy_expression

st1page:
refers FlinkSQL's watermark concept, change the watermark's `time_out` to `watermark_strategy_expression`. The PR
will be merged into #2 after being approved

### DIFF
--- a/rfcs/0002-watermark-filter.md
+++ b/rfcs/0002-watermark-filter.md
@@ -228,7 +228,7 @@ SELECT
   lead(price, 1) over w as next_price,
   lead(order_time, 1) over w as next_order_time
 FROM 
-  WATERMARK(orders, order_time, INTERVAL '1' MINUTE), 
+  WATERMARK(orders, order_time, order_time - INTERVAL '1' MINUTE), 
 WHERE
   abs(next_price - price) > 10000
   AND next_order_time - order_time < INTERVAL '30' SECOND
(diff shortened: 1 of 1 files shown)


TennyZhuang:
#2

yuhao-su:
LGTM

rfcs/0002-watermark-filter.md
```diff
@@ -172,9 +172,12 @@ The core idea of Exchange is **Prop 1.1.** We need also to align the watermarks,
 We can discuss the syntax later, but now we can implement that as a table function and expose the watermarkFilter’s behavior directly.
 
 ```plain
-WATERMARK(orders, time_column, timeout: interval)
+WATERMARK(orders, time_column, watermark_strategy_expression)
 ```
 
+And the concept `watermark_strategy_expression` refers [FlinkSQL](https://nightlies.apache.org/flink/flink-docs-master/docs/dev/table/sql/create/#watermark). The expression is evaluated for every record and update the watermark if the result greater than the current watermark.
+The new synatx is more friendly for flinkSQL user and offers user methods defining their strategy
```
BugenZhao, line 179:
```suggestion
The new syntax is more friendly for flinkSQL user and offers user methods defining their strategy.
```

rfcs/0002-watermark-filter.md
```diff
@@ -235,6 +238,52 @@ WINDOW w AS (
 )
 ```
 
+And `watermark_strategy_expression` offers potential to make user define their own strategy. Some following case is fancy and not very well-defined. You can treat them as a pseudocode which we might implement in furture
```
BugenZhao, line 241:
```suggestion
And `watermark_strategy_expression` offers potential to make user define their own strategy. Some following case is fancy and not very well-defined. You can treat them as a pseudocode which we might implement in future.
```

PR #10, opened 2022-11-03
Auto Execution Mode Selection

chenzl25:


rfcs/0010-auto-execution-mode-selection.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Design
+
+Since we don't have any statistics about table/mv, we are unable to calculate a meaningful cost for queries, so we focus on the oltp workload read pattern to decide what queries should run in local mode.
+
+### Query should run in local mode:
+
+- Single table primary key lookup. we can check our physical plan to ensure whether there is only one table scan in the plan tree and this table scan contains a point lookup scan range.
+- Single table two side bounded range scan. we can check our physical plan to ensure whether there is only one table scan in the plan tree and this table scan contains a two side bounded scan range.
+- Single table Index lookup. we can check our logical plan and physical plan to ensure there is only one table scan in the logical plan tree and there is a lookup join(non covering index) or there is a table scan in the physical plan tree(covering index).
```
liurenjie1024, line 26: Another case is expression only sql, for example `select 1`

liurenjie1024:
We should make ```set query_mode=auto``` default ？

rfcs/0010-auto-execution-mode-selection.md
```diff
@@ -0,0 +1,94 @@
...
+      "SELECT c FROM sbtest%u WHERE id BETWEEN ? AND ?",
+      t.INT, t.INT},
+   sum_ranges = {
+      "SELECT SUM(k) FROM sbtest%u WHERE id BETWEEN ? AND ?",
+       t.INT, t.INT},
+   order_ranges = {
+      "SELECT c FROM sbtest%u WHERE id BETWEEN ? AND ? ORDER BY c",
+       t.INT, t.INT},
+   distinct_ranges = {
+      "SELECT DISTINCT c FROM sbtest%u WHERE id BETWEEN ? AND ? ORDER BY c",
```
fuyufjh, line 57: This query seems hard to say it's TP or AP unless `BETWEEN ? AND ?` is proved to be a small range. I think we can exclude it from the test.

liurenjie1024 (approved):

rfcs/0010-auto-execution-mode-selection.md
fuyufjh: Actually, `select` any expression without tables?

chenzl25:
> We should make `set query_mode=auto` default ？

Yes.

PR #11, opened 2022-11-03
RFC: Coarse grained resource management for batch query engine

liurenjie1024:


fuyufjh:
LGTM. 

It need extra carefulness to rely on `catch_unwind` but I think it's acceptable, because all the execution in batch engine are wrapped as task, which seems to make it easier.

BowenXiao1999:
LGTM

fuyufjh:
Will this apply for both local & distribution execution mode?

liurenjie1024:
> Will this apply for both local & distribution execution mode?

Yes.

skyzh:
`catch_unwind` requires all future to impl `UnwindSafe`. But AFAIK, most of the futures in RisingWave doesn't guarantee this. Would you please try add `catch_unwind` on some random futures in the batch engine and see if it compiles?

liurenjie1024:
> `catch_unwind` requires all future to impl `UnwindSafe`. But AFAIK, most of the futures in RisingWave don't guarantee this. Would you please try add `catch_unwind` on some random futures in the batch engine and see if it compiles?

Yes, but I realized that we don't need `catch_unwind` since we can get spawned future's `JoinHandle` and the error will return to us through the `JoinHandle`'s `await` method.

An example can be found [here](https://github.com/risingwavelabs/risingwave/pull/6220)

fuyufjh:
After thinking twice I disagree with the design. The most critical problem is that when memory is used up, you can never know which thread it will break.

liurenjie1024:
> After thinking twice I disagree with the design. The most critical problem is that when memory is used up, you can never know which thread it will break.

Yes,  I'm afraid this will not be as easy as I think. We need another mechanism for this.

liurenjie1024:
I'll propose another rfc and let's close this for now.

liurenjie1024:
> All we have agreed is that we had better choose some simple and reliable solution. `panic` is simple but not reliable, so we excluded it. The simplest way I can tell is to set a limitation for each batch operator/stage and let them return an `Error` and then interrupt the whole query (or spill out, in the future).
> 
> There are a bunch of ways to limit the memory usage in operator/stage-level, for example, if the operator's data structures allow to overload the allocator, we can get the actual memory consumption; otherwise, we can simply set an upper bound for the number of the buffered rows, or estimatied size of buffered rows, which seems a little bit better than row count. These approaches are certainly naive and crude, but quite practical.

In fact, I have mentioned these approaches in alternatives section, and I didn't use it since it involves much more engineering effort. But I agree that panic is too simple to catch this error. I've updated the doc to use the alternative section.

> 
> Anyway, we don't actually want users to run rather complex queries on RisingWave. I tend to consider our batch engine as _primarily_ being designed for serving, and _sometimes_ for trying the queries before creating MView. 

I don't really want to argue about user requirements here, it's meaningless without investigation.

rfcs/0009-batch-query-resource-management.md
```diff
@@ -0,0 +1,69 @@
...
+
+2. Collections not supporting allocation api. In this case we need to estimate the memory usage of these containers.
+
+3. Data chunks.
+
+Notice that we can't simply use atomic variable here, we should use [hytra](https://docs.rs/hytra/latest/hytra/) here to accumulate memory usage.
+
+## Unresolved questions
+
+Why we need to collect batch tasks memory usage rather using process memory usage directly, since according to current streaming memory management design, we may starving batch tasks. Also this makes thing unpredictable and diffcult to explain to users.
```
StrikeW, line 61: > Why we need to collect batch tasks memory usage rather using process memory usage directly

IIUC, the condition to kill batch tasks is `other_mem_usage + batch_task_mem > system_safe_mem`. But to get the value of `other_mem_usage`we still need to depend on the value of `system_mem_usage`, which is the process memory usage I think. So I don't understand what is the difference here.

liurenjie1024: Yes, you are right. In fact, what I mean here is that we need to identify `batch_task_mem` rather a simpler solution like `system_mem_usage > batch_mem_threshould`, which may make things unpredictable and difficult to explain.

liurenjie1024:
> 1. The limitation to batch queries' memory is too relaxed. We need a hard limit.

We just need to lower the `system_safe_mem` to `batch_task_mem + 30%`, then we will start to kill task when it occupies too much memory.

> 2. Collecting memory usage by injecting allocators is not feasible.

I've listed solution to this, for collections without allocation api, we have to use estimation. But for the case `HashMap`, we can replace it with `https://docs.rs/hashbrown/latest/hashbrown/struct.HashMap.html`.

BowenXiao1999:
Indeed, according to current design, the limitation to batch is not a hard limit -- When batch memory exceeds `batch_task_mem`, we only stop accepting new task and running tasks may still grab more memory from streaming. I'm not sure how exact the situation will be, just points out some possibilities that may hurts streaming. 

So there is a trade off: If batch once exceed the threashold, 1) we can kill it, then we can ensure streaming safety. 2) If we just pending new task, old tasks may grow larger (if we do not have too much MPP tasks, I think this condition is rare). Frequent kill cause batch task unstable (hurt user's experience) --- it's maybe not better than just setup hard limit on specific memory-heavy executor e.g. `IndexJoin` or `HashJoin` join entry's number.


For using Allocator to collect Collections, Streaming also uses `Allocator` + estimmated size before. Now they disabled it and switched back to use jemalloc stats and I'm not sure the pros and cons on this `Allocator` implementation detail. Seems like the latency is high. cc @yuhao-su Are there some experience to share?

liurenjie1024:
It seems that there are still misunderstanding here, I'll schedule another meeting to discuss this.

liurenjie1024:
> Indeed, according to current design, the limitation to batch is not a hard limit -- When batch memory exceeds `batch_task_mem`, we only stop accepting new task and running tasks may still grab more memory from streaming. I'm not sure how exact the situation will be, just points out some possibilities that may hurts streaming.

In fact, if once task's mem keeps growing, `system_mem_usage`(`batch_task_usage + other_mem_usage`) will exceed `system_safe_mem`, then we will start killing batch task. I don't deny that it may hurt streaming, but this solution is to avoid exhausting memory, rather  perfect resource isolation. Memory is not the only thing that should be isolated, cpu, disk io, network all may hurt each other. In fact when user runs olap memory along with streaming, it's assumed that user accept the possibility that they may affect each other. If user wants perfact, they need a dedicated serving cluster.

> So there is a trade off: If batch once exceed the threashold, 1) we can kill it, then we can ensure streaming safety. 2) If we just pending new task, old tasks may grow larger (if we do not have too much MPP tasks, I think this condition is rare). Frequent kill cause batch task unstable (hurt user's experience) --- it's maybe not better than just setup hard limit on specific memory-heavy executor e.g. IndexJoin or HashJoin join entry's number.

Limiting entry number of memory heavy executor is impractical since we can't adjust it dynamically according join entry size, available memory in system, concurrent running queries.

> For using Allocator to collect Collections, Streaming also uses Allocator + estimmated size before. Now they disabled it and switched back to use jemalloc stats and I'm not sure the pros and cons on this Allocator implementation detail. Seems like the latency is high. cc @yuhao-su Are there some experience to share?

In fact I've confirmed with @yuhao-su `Allocator` with collections doesn't work because `AtomicUsize`'s performance is quite bad with contention, and `TrAddr` can resolve this.

yuhao-su:
>  Seems like the latency is high.

This is caused by the way we estimate the size of a collection. In the current implementation, we scan the entire collection to get the size of all items. Thus the time cost can be hight.

fuyufjh:
> In fact, if once task's mem keeps growing, system_mem_usage(batch_task_usage + other_mem_usage) will exceed system_safe_mem, then we will start killing batch task. I don't deny that it may hurt streaming, but this solution is to avoid exhausting memory, rather perfect resource isolation

The problem is that streaming doesn't follow this 2-threshold (soft + hard) design. In contrast, streaming will always evict some entries as long as the **global** memory usage is almost used up. 

Because both streaming and batch engine measure **global memory usage**, they are actually contending for memory. What we hoped is on the contrary - we want to reserve resources for streaming.

An concrete consequence of this problem is as described in my example: 

> A typical and usual case is that one complex batch query was issued by the user and then it consumes almost all the memory. According to the solution described in Limit number of memory used in batch query engine, the memory usage will keep growing; meanwhile, the streaming engine will detect the insufficient memory and keeps evicting memory by increasing the LRU watermark.

liurenjie1024:
> Because both streaming and batch engine measure **global memory usage**, they are actually contending for memory. What we hoped is on the contrary - we want to reserve resources for streaming.

Yes, but in fact streaming has two threshold: low_global_usage and high_global_usage. If memory exceeds `low_global_usage`, it starts evict memory.  To solve the problem you mentioned, we just need to set `system_safe_mem` usage to a little lower than `low_global_usage`, then we will start to kill olap task before streaming begin to evict memory.  The reason I don't set another bar for batch memory is that when the system is idle, say there is no much streaming task in system, we want to adhoc query to use memory as much as possible. And this case happens a lot when user uses a simple single node deployment to do some experiments.

skyzh:
> I've listed solution to this, for collections without allocation api, we have to use estimation. But for the case HashMap, we can replace it with https://docs.rs/hashbrown/latest/hashbrown/struct.HashMap.html.

Indeed, Rust std's hashmap implementation is hashbrown 🤣

StrikeW:
Random idea: In theory, batch query always read a bounded number of rows. Can we estimate the amount of memory for a query from some table/index/storage statistics? Then simply reject to run it if exceed the threshold.

liurenjie1024:
> Random idea: In theory, batch query always read a bounded number of rows. Can we estimate the amount of memory for a query from some table/index/storage statistics? Then simply reject to run it if exceed the threshold.

It's difficult to estimate it, especially in distributed environment.

liurenjie1024:
Conclusion:
1. We will use one task to control the logic of evicting streaming cache and killing batch olap task
2. We prefer estimating memory usage over adding allocator trait, only use allocator when possible.

BowenXiao1999:
A little miss from yesterday's meeting, some questions:

> We will use one task to control the logic of evicting streaming cache and killing batch olap task

Do we still reject new tasks when after line1 and killing when it exceeds line2. Or just killing after line1. IIUC, we need to implement this logic in `lru_manager::run` .

> We prefer estimating memory usage over adding allocator trait, only use allocator when possible

IIUC, we need to add `EstimatedSize` for HashMap, BTreeMap, DataChunk struct like https://github.com/risingwavelabs/risingwave/blob/52b0b8ee2004b15bfc6413203a2e4b10cd77adb1/src/stream/src/executor/managed_state/join/join_entry_state.rs#L97-L102,

and collect all this info in a Batch Context (CN level). Read this as batch mem usage when we lru_manager decide to evict memory?

fuyufjh:
> > We will use one task to control the logic of evicting streaming cache and killing batch olap task

> Do we still reject new tasks when after line1 and killing when it exceeds line2. Or just killing after line1. IIUC, we need to implement this logic in lru_manager::run .

I would definitely +1 for using one coroutine. But for the policy part, I would like to keep it simple and easy to understand. The first and foremost objective is that: it should be configurable correctly by users themselves.

The original policy proposed by Renjie is too complex for me. I am proposing another policy.

Anyway, this is merely the policy part, it could be changed easily.

liurenjie1024:
> Do we still reject new tasks when after line1 and killing when it exceeds line2. Or just killing after line1. IIUC, we need to implement this logic in lru_manager::run .

Yes, we should put this logic in one coroutine, and then all @fuyufjh 's concern is gone and meaningless. I think keep policy in this proposal is both flexible and easy to understand. We can fully utilize memory when workload is not heavy while giving up memory when workload is busy.

> > We prefer estimating memory usage over adding allocator trait, only use allocator when possible

> IIUC, we need to add EstimatedSize for HashMap, BTreeMap, DataChunk struct like https://github.com/risingwavelabs/risingwave/blob/52b0b8ee2004b15bfc6413203a2e4b10cd77adb1/src/stream/src/executor/managed_state/join/join_entry_state.rs#L97-L102,
and collect all this info in a Batch Context (CN level). Read this as batch mem usage when we lru_manager decide to evict memory?

Yes. The estimated size is used to estimate what can't be estimated in allocator. Let's use `Vec<DataChunk>` as an example, it has two parts: memory managed by `Vec` and memory managed by `DataChunk`, we can use allocator to collector `Vec`'s memory, and estimate each `DataChunk`'s memory.

BowenXiao1999:
Seems like another point have no discussed:

How should exeuctor report their bytes usage to task level context? Streaming use metrics and update it when barrier comes, as illustrated in https://github.com/risingwavelabs/risingwave/blob/2d873838afb7b50af40c5c44c8e1c52a41c11ae7/src/stream/src/executor/hash_join.rs#L710-L714 
(though it is disabled now cuz the latency is high)

I think for batch we can
1) add esitimated size in `Executor` trait.
2) update the number when `.next()` is called. Root executor will need to collect size recursively to get task level number. https://github.com/risingwavelabs/risingwave/blob/2d873838afb7b50af40c5c44c8e1c52a41c11ae7/src/batch/src/task/task_execution.rs#L428-L432 

Because if we use a background process to poll, it will become more complex i guess..

yuhao-su:
> Seems like another point have no discussed:
> 
> How should exeuctor report their bytes usage to task level context? Streaming use metrics and update it when barrier comes, as illustrated in https://github.com/risingwavelabs/risingwave/blob/2d873838afb7b50af40c5c44c8e1c52a41c11ae7/src/stream/src/executor/hash_join.rs#L710-L714 (though it is disabled now cuz the latency is high)
> 

One possible solution can be something like #5932. We can do some easy modifications to register all collect memory stats to the lru manager.

fuyufjh (approved):
Shall we merge this?

PR #12, opened 2022-11-04
Add external function

TennyZhuang:
Signed-off-by: TennyZhuang <zty0826@gmail.com>

### DIFF
--- a/rfcs/0000-template.md
+++ b/rfcs/0000-template.md
@@ -1,0 +1,36 @@
+---
+feature: my_excited_feature
+authors:
+  - "TennyZhuang"
+start_date: "2022/10/24"
+---
+
+# My Excited Feature
+
+Please feel free to add or remove sections.
+
+## Summary
+
+Explain the feature in short.
+
+## Motivation
+
+Why are you want to introduce the feature?
+
+## Design
+
+Explain the feature in detail.
+
+## Unresolved questions
+
+* Are there some questions that haven't been resolved in the RFC?
+* Can they be resolved in some future RFCs?
+* Move some meaningful comments to here.
+
+## Alternatives
+
+What other designs have been considered and what is the rationale for not choosing them?
+
+## Future possibilities
+
+Some potential extensions or optimizations can be done in the future based on the RFC.


TennyZhuang:
@skyzh Can you ignore the HackMD bot in CLA checker?

rfcs/0012-external-function.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Future possibilities
+
+### Using another RPC framework
+
+The external function framework will not be public to users, so we use gRPC in our first version for simplicity. We can investigate other RPC frameworks later.
+
+### Using wasm-based implementation and embedding the VM into the kernel?
+
+A good idea, but currently, the capacity of the wasm is limited. For some CPU-bounded functions written in some languages without heavy runtime, it's a better solution for performance.
```
BugenZhao, line 140:
```suggestion
A good idea, but currently, the capability of the wasm is limited. For some CPU-bounded functions written in some languages without heavy runtime, it's a better solution for performance.
```
?

rfcs/0012-external-function.md
```diff
@@ -0,0 +1,140 @@
...
+```plain
+LogicalProject { exprs: [round($0)] }
+└─LogicalAsyncProject { exprs: [my_udf($0)] }
+  └─LogicalProject { exprs: [to_char($0)] }
+    └─LogicalScan { table: t, columns: [t.x] }
+```
+
+The solution can work well on `Project`, `Filter`, and `Join`, but failed on [the filter clause](https://medium.com/little-programming-joys/the-filter-clause-in-postgres-9-4-3dd327d3c852). There are also many other corner cases since expressions are used everywhere in PostgreSQL.
+
+Also, this is not an elegant solution. For the frontend part, `Project` and `ProjectSet` are the same.
```
BugenZhao, line 130:
```suggestion
Also, this is not an elegant solution. For the frontend part, `Project` and `ProjectAsync` are the same.
```
?

BowenXiao1999 (commented):
1. Can you also provide more context why the interface should be async?
2. How to understand cross-chunk concurrency? Do you mean if udf1 and udf2 both need chunk A, then they shared read this chunk?

BugenZhao:
> 2\. How to understand cross-chunk concurrency?

Take `Project` as an example, if the UDF is I/O bounded, then we may want to process the input chunks in a pipeline manner by `buffer`ing them to be concurrent.

skyzh:
Do we allow external functions in places other than projection? (e.g., filter?)

TennyZhuang:
> Do we allow external functions in places other than projection? (e.g., filter?)

Yes, I hope everywhere.

skyzh:
Before introducing external functions to the system, we should also consider deterministic problems.

Considering the following query:

```
select * from (select udf(x) as x from table) where x > 10;
```

Suppose we get the following events from table:

```
INSERT 1
DELETE 1
```

If udf is not deterministic, suppose it gets 1 -> 11 for the first message, and 1 -> 10 for the second message.

Many things will go wrong. The filter will emit:

```
INSERT 1
```

which ignores the second delete.

One feasible solution is to have a special kind of projection called MaterializeProjection, where we materialize data by their pk, so that UDF will only be computed once at the first time when the generated event goes into the system.

skyzh:
That is to say, `MaterializeProjection` is a cache for UDF, whereas given one input we will always produce the same output for a given UDF.

TennyZhuang:
> Before introducing external functions to the system, we should also consider deterministic problems.
> 
> Considering the following query:
> 
> ```
> select * from (select udf(x) as x from table) where x > 10;
> ```
> 
> Suppose we get the following events from table:
> 
> ```
> INSERT 1
> DELETE 1
> ```
> 
> If udf is not deterministic, suppose it gets 1 -> 11 for the first message, and 1 -> 10 for the second message.
> 
> Many things will go wrong. The filter will emit:
> 
> ```
> INSERT 1
> ```
> 
> which ignores the second delete.
> 
> One feasible solution is to have a special kind of projection called MaterializeProjection, where we materialize data by their pk, so that UDF will only be computed once at the first time when the generated event goes into the system.

Good catch! I have no idea how to resolve that, how about only allow all UDFs on append-only stream?

TennyZhuang:
> * Can you also provide more context why the interface should be async?

There is an RPC call during the evalutaion.

skyzh:
Append-only stream should be fine. But this means that people can only do UDF before aggregation / TopN, which seems to be a common case in SQL. Not sure about the use case.

TennyZhuang:
A basic idea:

Users must specify their UDFs are deterministic explicitly, otherwise, the UDFs can only be applied on an append-only stream or be materialized implicitly.

rfcs/0012-external-function.md
```diff
@@ -0,0 +1,163 @@
...
+### External function protocol
+
+We will add a simple protocol for external function servers to implement:
+
+```protobuf
+message BatchExternalUdfRequest {
+    repeated bytes columns = 1;
+}
+
+message BatchExternalUdfResponse {
```
liurenjie1024, line 34: How is the user expected to report error?

TennyZhuang: Not sure, but we can't handle the error elegantly, so just leave it later.

rfcs/0012-external-function.md
```diff
@@ -0,0 +1,163 @@
...
+We will introduce a gRPC-based external function framework in the kernel part. **The framework is private to users and may change later**.
+
+## Motivation
+
+The RFC is a preparation RFC for external UDFs with heavy runtime or complicated environments.
+
+The RFC will not answer the following questions:
+
+* How to define the UDFs?
+* How to execute the UDFs?
```
fuyufjh, line 21: Hmmm, but how could you do testing without these?

rfcs/0012-external-function.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Design
+
+### External function protocol
+
+We will add a simple protocol for external function servers to implement:
+
+```protobuf
+message BatchExternalUdfRequest {
```
liurenjie1024, line 30: It would be better to rename it to `ArrowExternalUdfRequest`, the word batch seems confusing to me.

rfcs/0012-external-function.md
```diff
@@ -0,0 +1,163 @@
...
+    repeated bytes columns = 1;
+}
+
+message BatchExternalUdfResponse {
+    bytes result = 1;
```
BugenZhao, line 35: Should we allow multiple columns in the response as well?

rfcs/0012-external-function.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Design
+
+### External function protocol
+
+We will add a simple protocol for external function servers to implement:
+
+```protobuf
+message BatchExternalUdfRequest {
+    repeated bytes columns = 1;
```
fuyufjh, line 31: What's the internal encoding of input/output columns?

rfcs/0012-external-function.md
```diff
@@ -0,0 +1,163 @@
...
+authors:
+  - "TennyZhuang"
+start_date: "2022/11/04"
+---
+
+# External Function Framework
+
+## Summary
+
+We will introduce a gRPC-based external function framework in the kernel part. **The framework is private to users and may change later**.
```
liurenjie1024, line 12: If we want to be compatible with arrow, why not use arrow flight rpc? https://arrow.apache.org/docs/format/Flight.html#

rfcs/0012-external-function.md
fuyufjh: How about defining such a syntax

```
create external function
```

https://docs.snowflake.com/en/sql-reference/sql/create-external-function.html

rfcs/0012-external-function.md
fuyufjh: If we want users to decode/encode a column, I think it's better to use some open format like Arrow or Avro. Otherwise we have to provide SDKs for all languages.

TennyZhuang: Yes, we prefer to use arrow, but we’ll try to implement it directly in our array implementation.

rfcs/0012-external-function.md
TennyZhuang: No objection

rfcs/0012-external-function.md
TennyZhuang: We don't use an arrow format library and write it manually, so it's hard for us to use a flight library directly, and we don't want to waste time for that.

TennyZhuang: Anyway, it's acceptable to me, but I don't want to investigate it now.

rfcs/0012-external-function.md
TennyZhuang: Currently unnecessary, pg will also pack it in struct format.

rfcs/0012-external-function.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Future possibilities
+
+### Using another RPC framework
+
+The external function framework will not be public to users, so we use gRPC in our first version for simplicity. We can investigate other RPC frameworks later.
+
+### Using wasm-based implementation and embedding the VM into the kernel?
+
+A good idea, but currently, the capability of the wasm is limited. For some CPU-bounded functions written in some languages without heavy runtime, it's a better solution for performance.
```
twocode, line 163: > the capability of the wasm is limited

Can you give some examples? 🤔

liurenjie1024: For me there are two points:
1. Currently there is no stable solution to convert java code to wasm
2. It seems that wasi is not fully supported.

wangrunji0408:
Can we merge this RFC? as most of its proposals have been implemented now.

TennyZhuang:
> Can we merge this RFC? as most of its proposals have been implemented now.

You can give an approval.

wangrunji0408 (approved):

PR #13, opened 2022-11-04
RFC: Use Backfill To Let Mv On Mv Stream Again

chenzl25:
image


hzxa21:
Note that regardless of what approach we take to optimize MV creation, we still need to have some kind of backpressure because we cannot assume that the downstream MV's throughput (for example, it is more complex) can catch up with upstream MV.

chenzl25:
> FYI, another way to reduce memory usage and accelerate MV creation is prorviding an interface to scan epoch diff from storage. We discussed it long time ago. See https://singularity-data.quip.com/Oaz4Ajo6Stu8/RFC-MView-granularity-Partial-Checkpoint-Based-on-MView-Epoch-ReplayWIP cc @wenym1

It seems epoch diff needed to pin the old epoch until the end of mv creation, but actually we don't have to based on this RFC.

chenzl25:
> Note that regardless of what approach we take to optimize MV creation, we still need to have some kind of backpressure because we cannot assume that the downstream MV's throughput (for example, it is more complex) can catch up with upstream MV.

Sure, the worst case of this RFC is we can't advance the backfill progress and keep forwarding the upstream update (which pk < `current_pos`) to down stream.

hzxa21:
> > FYI, another way to reduce memory usage and accelerate MV creation is prorviding an interface to scan epoch diff from storage. We discussed it long time ago. See https://singularity-data.quip.com/Oaz4Ajo6Stu8/RFC-MView-granularity-Partial-Checkpoint-Based-on-MView-Epoch-ReplayWIP cc @wenym1
> 
> It seems epoch diff needed to pin the old epoch until the end of mv creation, but actually we don't have to based on this RFC.

To scan epoch diff, we don't need to "pin" the old epoch like pinning a snapshot and delaying compaction. We just need to avoid deleting L0 SST objects in S3 in the vacuum process. Basically treating L0 as a Log. The upside of this approach is it can be very efficient becuase it basically just simply scans the file one by one and can avoid excessive buffer. The downside of this approach is storage needs to provide a special `scan_diff(start_epoch, end_epoch)`, which may increase the complexity. Btw, I just want to bring back the previous thoughts since it is somehow related and I am not agaist this RFC.

0013-use-backfill-to-let-mv-on-mv-stream-again.md
```diff
@@ -0,0 +1,51 @@
...
+
+There is a paper *Online, Asynchronous Schema Change in F1* introduced by google used to solve the online schema change problem of distributed system. They use online schema change to solve the creating index problem which somehow is similar to our creating index case or even our creating materialized view case. Online schema change use a backfill operation to backfill all missing key-value pairs corresponding to the new element and this backfill operation can run concurrently with any upstream update. This idea can be used to our system. In this way, we can eliminate the unbounded buffer of upstream update.
+
+## Design
+
+We use creating index as a example to illustrate the idea, however, we can apply it to create mv too.Suppose that we have a table T which contains 10,000 rows. We want to create an index on it. This index is actually also a materialized view. The backfill operations work like as following.
+
+### Backfill
+
+- Use a variable `current_pos` to keep track the position (pk) we have backfilled which initialized as 0.
```
skyzh, line 24: Does this work on a snapshot or we can move to the next snapshot?

0013-use-backfill-to-let-mv-on-mv-stream-again.md
```diff
@@ -0,0 +1,51 @@
...
+- We select (`current_pos`, `current_pos` + `batch_size`) from table T and insert them to the index. e.g. `batch_size` can be 1000.
+- Repeatedly backfill next batch after current batch to the index.
+- We can forward any update stream from table T to the index if its pk is lower than the `current_pos` concurrently, otherwise, igonre the update.
```
skyzh, line 27: How to handle deletes?

0013-use-backfill-to-let-mv-on-mv-stream-again.md
skyzh: I assume pk is sorted in an ascend way. If the first pk is "b" scanned by the backfill, the current_pos is "c", and there's an update from "a" (which is smaller than current_pos), what should we do?

chenzl25: > I assume pk is sorted in an ascend way. If the first pk is "b" scanned by the backfill, the current_pos is "c", and there's an update from "a" (which is smaller than current_pos), what should we do?

Forward update "a" to the downstream because it is smaller than current_pos. Because if current_pos is "b", we can ensure that any row with pk smaller than "b" have already been sent to downstream.

0013-use-backfill-to-let-mv-on-mv-stream-again.md
chenzl25: How to handle deletes is the same as updates or inserts. The only thing we need to take care is comparing stream message pk with `current_pos `.

- Forward message to downstream if its pk is smaller than `current_pos`.
- Ignore message if its pk is bigger than `current_pos` + `batch_size`.
- Apply message to buffered data if its pk is within (`current_pos`, `current_pos` + `batch_size`)

BugenZhao:
Since we don't need to buffer any records or persist them in a WAL manner with this design, I guess we can even do scaling on creating mviews by persisting the current position per vnode! 🥵

One of the advantages of this design is that the changes to the system are bounded in a single `Chain`(V3🤣) operator implementation, which is really decoupled and easy to maintain.

0013-use-backfill-to-let-mv-on-mv-stream-again.md
```diff
@@ -0,0 +1,51 @@
...
+
+We can save `current_pos` to the state and use it to help us recover after failover. It is useful if we need to create a mv on a large existing mv. It may takes a long time to create and if any error happen during creating, we can recover from the last position we record to save expensive work have been done before.
+
+
+## Unresolved questions
+
+
+## Future possibilities
+
+Each iteration of backfill need to immediately select next `epoch` data from the upstream, however, we need to wait epoch until the barrier go through the whole DAG now. Optimization can be used to speed up the wait epoch.
```
BugenZhao, line 50: Agree, or we have to stall most of the time.

Actually, the current scheduler ensures that `Chain` will only access the local records, so we don't even need to wait for the initial epoch for `RearrangedChain`. But it'll be better if the storage realizes this and hides the details in the `wait_epoch`. Will local state store help this? cc @hzxa21 @hzxa21

liurenjie1024:
Can we apply this is complex mviews? Creating index is somehow a simple case.

chenzl25:
> Can we apply this is complex mviews? Creating index is somehow a simple case.

Yes it can. I just use creating index as an example to illustrate for simplicity, but it is applicable to complex mv creation too.

BugenZhao:
The only assumption I can come up with to break is the ordering, while the `Chain` v1 and v2 keep the ordering as "snapshot first, then upstream". I'm not sure whether this'll be a problem, like for watermark? 🤔

rfcs/0013-use-backfill-to-let-mv-on-mv-stream-again.md
```diff
@@ -0,0 +1,69 @@
...
+<img width="970" alt="image" src="https://user-images.githubusercontent.com/9352536/200274858-7880302a-952e-4f39-9c85-0f59fe9d4233.png">
+
+We use creating index as an example to illustrate the idea, however, we can apply it to create mv too. Suppose that we have a table T which contains 10,000 rows. We want to create an index on it. This index is actually also a materialized view. The backfill operations work as follows.
+
+### Backfill
+
+Create a backfill operator which is similar to the Chain operator.
+
+- Use a variable `current_pos` to keep track of the position (pk) we have backfilled which is initialized as 0.
+- We select rows with pk between (`current_pos`, `current_pos` + `batch_size`) from table T into buffer. e.g. `batch_size` can be 1000.
```
fuyufjh, line 29: Actually, it scans from PK `current_pos` for at most 1000 elements

fuyufjh:
LGTM!

st1page (approved):
Would we merge now because we have finished it?

PR #14, opened 2022-11-07
RFC: Dedicated Serving Cluster

fuyufjh:
Preview: https://github.com/risingwavelabs/rfcs/blob/eric/dedicated-serving-cluster/rfcs/0014-dedicated-serving-cluster.md

------------

2022 Nov 9: Updated the docs with

- Add the section "Non-Goals" to clarify the problem.
- Better approaches to non-checkpoint reads. Thanks to @BowenXiao1999 and @hzxa21.

### DIFF
--- a/rfcs/0014-dedicated-serving-cluster.md
+++ b/rfcs/0014-dedicated-serving-cluster.md
@@ -13,7 +13,8 @@ Run a dedicated cluster to serve batch queries.
 
 ## Motivation
 
-- **Different traffic pattern** of batch & streaming. Batch queries, especially the high-concurrent point queries (aka. “online serving”), requires millisecond-level low latency and high availability, while streaming jobs focus more on high throughput and cost efficiency.
+- **Different traffic pattern**. Batch queries, especially the high-concurrent point queries (aka. “online serving”), requires millisecond-level low latency and high availability, while streaming jobs focus more on high throughput and cost efficiency.
+- **Different access pattern on storage** of batch & streaming. Batch queries are issued by services on user side, while streaming operators always handle the changes from upstreaming system. It's possible that they have quite different distribution of access aka. hotspot. 
 - In some extreme use cases, the QPS on serving side could be 100K or even higher, which may result in severe **resource contention** (CPU, IOPS, etc.) against streaming jobs.
 
 To address these problems, here we propose to run a dedicated cluster to serve batch queries (hereafter referred to as “serving cluster”). Meanwhile, data source, sink and maintenance of MViews are still done in the primary streaming cluster.
--- a/rfcs/0014-dedicated-serving-cluster.md
+++ b/rfcs/0014-dedicated-serving-cluster.md
@@ -1,0 +1,75 @@
+---
+feature: dedicated-serving-cluster
+authors:
+  - "Eric Fu"
+start_date: "2022/11/07"
+---
+
+# Dedicated Batch Serving Cluster
+
+## Summary
+
+Explain the feature in short.
+
+## Motivation
+
+- **Different traffic pattern** of batch & streaming. Batch queries, especially the high-concurrent point queries (aka. “online serving”), requires millisecond-level low latency and high availability, while streaming jobs focus more on high throughput and cost efficiency.
+- In some extreme use cases, the QPS on serving side could be 100K or even higher, which may result in severe **resource contention** (CPU, IOPS, etc.) against streaming jobs.
+
+To address these problems, here we propose to run a dedicated cluster to serve batch queries (hereafter referred to as “serving cluster”). Meanwhile, data source, sink and maintenance of MViews are still done in the primary streaming cluster.
+
+Additionally, this should be an **optional** feature. Some simple use cases, like dashboard, may not need to support such high concurrency. Besides, free tier RisingWave Cloud users are definitely not its target users.  
+
+## Design
+
+Decoupling barrier and checkpoint (reference: [Support barrier-checkpoint decoupling](https://singularity-data.quip.com/QLOaAegyeApR/Support-barrier-checkpoint-decoupling)) allows us to decouple the streaming barrier from the expensive checkpointing process, but also brings more complexity to us.  
+
+### Read Checkpointed Snapshots
+
+In the first stage, we can simply boot up several compute nodes without scheduling any streaming tasks on them. The meta service pushes the new storage versions to the compute nodes as usual, therefore, whenever a new storage version generated, the compute node will be informed.
+
+In the catalog, beside the existing `vnode -> parallel unit` mapping, we have to maintain another mapping of serving cluster for each table & materialized view. Below is an example, where “PU” refers to Parallel Unit. Apparently, this mapping should be updated once nodes joined or leaved the serving cluster. 
+
+```jsx
+Main Cluster: 
+- PU#1: vnodes [0, 1]
+- PU#2: vnodes [2, 3]
+- PU#3: vnodes [4, 5]
+
+Serving Cluster: 
+- PU#1: vnodes [0, 1, 2]
+- PU#2: vnodes [3, 4, 5]
+```
+
+High availability can be achieved by allocating multiple Parallel Units for each vnode. For example, given 3 nodes and `replicas = 2`,
+
+```jsx
+Serving Cluster: 
+- PU#1: vnodes [0, 1, 2, 4]
+- PU#2: vnodes [3, 4, 5, 1]
+- PU#3: vnodes [0, 2, 3, 5]
+```
+
+### Read Non-Checkpointed Snapshots
+
+In the current design, non-checkpointing barriers are much more frequent than checkpointing barriers, so it provided both freshness and lower IO cost. However, these non-checkpointed updates are only kept in memory, more specifically, in the `StateStore` of the `MaterializedExecutor` with corresponding vnode.
+
+Here we propose to **store** these barrier-level changes of non-checkpoint barriers to shared storage **temporarily**. These immutable MemTable SSTs will only be temporarily placed in L0, and these SST files can be dropped as long as next checkpoint completes.
+
+To make it simpler, we can reuse the concept “compaction group”. A compaction group can be tagged as checkpointing on every barrier. Table_ids in such compaction groups will upload its immutable MemTable on non-checkpointing barriers, and a “serving-only” version will be created.
+
+Note that serving-only versions will not affect the recovery process. On recovery, all nodes should roll back the latest completed checkpoint like before.
+
+## Unresolved questions
+
+None
+
+## Alternatives
+
+**How about streaming changes to serving nodes?**
+
+This is possible. However, since we hope to allow both cluster to have different SLA, and furthermore, can be scaled independently, the streaming here may be complicated than imagination. 
+
+## Future possibilities
+
+None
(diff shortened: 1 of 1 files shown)


BowenXiao1999 (commented):
1. Just curious, are there some numbers or benchmark showing that we need this cluster to handle high concurrency? The resoruce management proposed by @liurenjie1024 seems can also mitigate this problem. 
2. For un-checkpointed changes, if we still flush, what's the difference compared to checkpointed changes?
3. Can we inform Serving Cluster with the vnode mapping of primary streaming cluster so scheduler can fire some tasks to directly get un-checkpointed data in memory?

Rest LGTM

fuyufjh:
> 1. Just curious, are there some numbers or benchmark showing that we need this cluster to handle high concurrency? The resoruce management proposed by @liurenjie1024 seems can also mitigate this problem.

No. I don't think it's designed to achieve higher concurrency than now, so benchmarks cannot show anything. As for resource management, Renjie's proposal limits the memory usage, but that's still far away from physical resource isolation.

> 2. For un-checkpointed changes, if we still flush, what's the difference compared to checkpointed changes?

Only states of materialized views & tables are flush. I assume it's smaller than the operator state.

> 3. Can we inform Serving Cluster with the vnode mapping of primary streaming cluster so scheduler can fire some tasks to directly get un-checkpointed data in memory?

Sounds like a good idea!

rfcs/0014-dedicated-serving-cluster.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Design
+
+Decoupling barrier and checkpoint (reference: [Support barrier-checkpoint decoupling](https://singularity-data.quip.com/QLOaAegyeApR/Support-barrier-checkpoint-decoupling)) allows us to decouple the streaming barrier from the expensive checkpointing process, but also brings more complexity to us.  
+
+### Read Checkpointed Snapshots
+
+In the first stage, we can simply boot up several compute nodes without scheduling any streaming tasks on them. The meta service pushes the new storage versions to the compute nodes as usual, therefore, whenever a new storage version generated, the compute node will be informed.
+
+In the catalog, beside the existing `vnode -> parallel unit` mapping, we have to maintain another mapping of serving cluster for each table & materialized view. Below is an example, where “PU” refers to Parallel Unit. Apparently, this mapping should be updated once nodes joined or leaved the serving cluster.
```
BugenZhao, line 32: How do we determine and reschedule this partition mapping? Should we have some heuristic strategies to replicate the hot spot more, or we simply scale with the cluster?

fuyufjh: That belongs to optimizations. For now, I prefer to assume the traffic can always be well distributed to all vnodes (so no "hot" vnode)

hzxa21:
## Read Checkpointed Snapshots

I have the similar ideas. Just two more points to add:

- Serving cluster and the Streaming Cluster can have the same vnode mapping by default, which is more friendly to cache.
- If we only care about availability, not performance, the serving cluster vnode mapping can be served as a hint, instead of a requirement, for executor scheduling. In other words, if we lose all replicas of a vnode in the serving cluster, we can still serve query to this vnode in other serving CNs since the state is on S3 and the CN is stateless.

## Read ****Non-Checkpointed Snapshots****

Writing to S3 on non-checkpointed barrier may not be ideal because write-to-S3 can significantly increase the non-checkpoint barrier latency (and thus freshness) and if we can afford the latency increase, we can just do checkpoint on every barrier.

Therefore, I am mainly considering the approaches without persisting uncommitted states:

- Streaming (Push) changes to serving nodes: as you mentioned in the doc, I also think this is complicated, especially when considering who to send changes (executor or storage?) and how to handle failover/scaling (basically vnode mapping changes)
- Pull changes from streaming nodes (proposed by ): a pull-based approach seems simpler. A serving CN only needs to fetch and cache uncommitted state according to the streaming vnode mapping, which is similar to what we did to co-schedule batch and MV executor.

Note that when there is a streaming CN failure, the uncommitted states on batch CN still need to be cleared but this can be done asynchronously.

hzxa21:
> Can we inform Serving Cluster with the vnode mapping of primary streaming cluster so scheduler can fire some tasks to directly get un-checkpointed data in memory?

Exactly. 👍

rfcs/0014-dedicated-serving-cluster.md
```diff
@@ -0,0 +1,76 @@
...
+# Dedicated Batch Serving Cluster
+
+## Summary
+
+Run a dedicated cluster to serve batch queries.
+
+## Motivation
+
+- **Different traffic pattern**. Batch queries, especially the high-concurrent point queries (aka. “online serving”), requires millisecond-level low latency and high availability, while streaming jobs focus more on high throughput and cost efficiency.
+- **Different access pattern on storage** of batch & streaming. Batch queries are issued by services on user side, while streaming operators always handle the changes from upstreaming system. It's possible that they have quite different distribution of access aka. hotspot.
```
liurenjie1024, line 17: Another access pattern is range query.

liurenjie1024:
There are two concerns with this approach:
1. If batch query has different access pattern, then why still need consisten hash for mv?
2. The cost of whole system. Now we have a serving cluster which contains enough cache for reading, then when stream executor meets cache miss, why it needs to load it from s3 rather this cluster? According our current number, loading from a kv's latency is less than 1ms, while from s3 it takes 20+ ms.

BugenZhao:
> 1. If batch query has different access pattern, then why still need consisten hash for mv?

The problem is whether to apply `RangeExchange` in the whole streaming system.
- If so, then all mview and internal states should be range-based, which sounds okay except for the complexity of scaling.
- If not, i.e., the stream is still partitioned by consistent hash. Then MV-on-MV will be a problem: we have two read patterns for the MV data.

liurenjie1024:
> The problem is whether to apply RangeExchange in the whole streaming system.
>
> * If so, then all mview and internal states should be range-based, which sounds okay except for the complexity of scaling.
> * If not, i.e., the stream is still partitioned by consistent hash. Then MV-on-MV will be a problem: when ingesting the historical data, we need a way to shuffle the range partition to consistent hash again.

I don't understand why we need range exchange? For cache locality? In my opnion, cache only exists in memory. For example, we have following distribution in serving cluster:

```
Node: [1, 2],
Node: [3, 4],
Node: [5, 6],
```

But in stateful streaming executor, we can still have hash based distribution:

```
Node: [1, 3, 5],
Node: [2, 4, 6]
```

hzxa21:
> 2. The cost of whole system. Now we have a serving cluster which contains enough cache for reading, then when stream executor meets cache miss, why it needs to load it from s3 rather this cluster? According our current number, loading from a kv's latency is less than 1ms, while from s3 it takes 20+ ms.

I think the serving cluster proposed in this RFC only caches queryable states (i.e. materialized view states) so streaming executor can hardly benefit from it since streaming won’t read materialized view state (as for now).

If we introduce a storage cache layer, in other words, we have a separate layer on top of S3 to cache all kinds of committed states including internal state, then both batch and streaming executor can benefit from it. It is a trade-off between cost and performance. Regardless of whether we introduce it now, IMO the storage cache layer is orthogonal to the batch CN cluster since we won’t schedule batch executors on the storage cache layer (predicate push-down to storage may be doable though) .

liurenjie1024:
> I think the serving cluster proposed in this RFC only caches queryable states (i.e. materialized view states) so streaming executor can hardly benefit from it since streaming won’t read materialized view state (as for now).

For example, the following sql
```
create materialized view as 
select advId, sum(revenue), count(*) from sales group by advId
```
In this case when we have cache miss, it needs to read from materialized view state.

> If we introduce a storage cache layer, in other words, we have a separate layer on top of S3 to cache all kinds of committed states including internal state, then both batch and streaming executor can benefit from it. It is a trade-off between cost and performance. Regardless of whether we introduce it now, IMO the storage cache layer is orthogonal to the batch CN cluster since we won’t schedule batch executors on the storage cache layer (predicate push-down to storage may be doable though) .

I agree that having a separate cluster for serving layer is this RFC's main focus. But having some machine to host queries is trivial, the key is how the storage is distributed. Also why we won't schedule batch executors on storage cache layer? I think it would be necessary to coschedule batch query with storage cache to improve performance.

Little-Wallace:
According this RFC, should we split a small compaction-group to each CN even if the data size of which is only several GBs?

rfcs/0014-dedicated-serving-cluster.md
```diff
@@ -0,0 +1,76 @@
...
+
+Run a dedicated cluster to serve batch queries.
+
+## Motivation
+
+- **Different traffic pattern**. Batch queries, especially the high-concurrent point queries (aka. “online serving”), requires millisecond-level low latency and high availability, while streaming jobs focus more on high throughput and cost efficiency.
+- **Different access pattern on storage** of batch & streaming. Batch queries are issued by services on user side, while streaming operators always handle the changes from upstreaming system. It's possible that they have quite different distribution of access aka. hotspot. 
+- In some extreme use cases, the QPS on serving side could be 100K or even higher, which may result in severe **resource contention** (CPU, IOPS, etc.) against streaming jobs.
+
+To address these problems, here we propose to run a dedicated cluster to serve batch queries (hereafter referred to as “serving cluster”). Meanwhile, data source, sink and maintenance of MViews are still done in the primary streaming cluster.
```
StrikeW, line 20: Seems like we should run a dedicated batch engine on a separate cluster. IMHO, to solve these problems effectively, maybe we need to take a big step:  a separate batch engine plus a separate storage format.

rfcs/0014-dedicated-serving-cluster.md
fuyufjh: I agree there will be more scans in batch but I don't think they are all ordered by PK. 

On the other hand, distributing data in another way rather than vnode will be much more complicated, which means it cannot leverage the existing block-level caching. If you are thinking about a KV storage without vnode, why not just introducing a deployment of TiKV or Redis? I mean, otherwise you actually have to reinvent an TiKV/Redis to do it.

liurenjie1024: Range partition not only orders keys, but also helps pruning unnecessary scan, which is critical in our performance. I admit that things would be more complicated if we have new partitioning schema, but we can't rush code every time when customer requirements appears, our architecture should be general enough.

fuyufjh:
> According this RFC, should we split a small compaction-group to each CN even if the data size of which is only several GBs?

Didn't get the point. Why is it?

rfcs/0014-dedicated-serving-cluster.md
```diff
@@ -0,0 +1,96 @@
...
+
+#### Approach 2. Read non-checkpoint writes from S3
+
+Here we propose to **store** these barrier-level changes of non-checkpoint barriers to shared storage **temporarily**. These immutable MemTable SSTs will only be temporarily placed in L0, and these SST files can be dropped as long as next checkpoint completes.
+
+To make it simpler, we can reuse the concept “compaction group”. A compaction group can be tagged as checkpointing on every barrier. Table_ids in such compaction groups will upload its immutable MemTable on non-checkpointing barriers, and a “serving-only” version will be created.
+
+Note that serving-only versions will not affect the recovery process. On recovery, all nodes should roll back the latest completed checkpoint like before.
+
+#### Approach 3. Contantly streaming changes from primary cluster
```
TennyZhuang, line 79: Vote for approach 3, it's similar to share arrangement

TennyZhuang: In our previous design, the incremental updates to share arrangements will be broadcasted to different actors. The behavior is what we expected here.

liurenjie1024 (approved):

PR #15, opened 2022-11-07
RFC: Permit-based Back-pressure with Bidirectional gRPC Streaming

BugenZhao:
Signed-off-by: Bugen Zhao <i@bugenzhao.com>

### DIFF
--- a/rfcs/0015-permit-based-back-pressure.md
+++ b/rfcs/0015-permit-based-back-pressure.md
@@ -1,0 +1,116 @@
+---
+feature: permit-based-back-pressure
+authors:
+  - "Bugen Zhao"
+start_date: "2022/10/24"
+---
+
+# Permit-based Back-pressure with Bidirectional gRPC Streaming
+
+## Background
+
+The back-pressure is necessary in a streaming system to ensure the resources are not exhausted. Luckily, it’s natural and easy to implement back-pressure in RisingWave, as what we’re doing now:
+
+- **Local Exchange**
+We use a channel bounded with the chunk number to exchange. When the downstream actor processing slowly, the channel will be filled gradually, and the upstream will be blocked.
+- **Remote Exchange**
+We rely on the HTTP/2 flow control under gRPC. When the downstream actor processing slowly, the window of the HTTP/2 stream will be filled gradually, and the upstream will be blocked.
+
+## Problems
+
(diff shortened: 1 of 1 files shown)


BugenZhao:
Untitled 1
Untitled

skyzh:
Have you run any experiment against the permit-based back-pressure?

Assume now all remote / local channels have the same window size, say 2MB. The receiver will need at least 1 RTT to do the full cycle of requesting data when window is full: send permit, wait for response.

(1) Assume RTT = 1s for remote channel and 0ms for local channel.
(2) Assume window size = 10000 rows.
(3) Assume the downstream executor consumes 50000 rows / second.
(4) When there's some slots in the receiver window, the remote channel will need at least 1s to refill (send permit, wait for reply). However, the window size is not large enough, and it will be empty after 0.2s.
(5) In this situation, the back-pressure is still unfair, as the receiver window for remote exchange cannot always be refilled.

To ensure fair back-pressure, we will need to fill the window to be larger than downstream process speed * RTT; however, larger window size also means more data in one checkpoint.

We might need to think carefully about it.

BugenZhao:
Yes. We must assume the latency is not that high and the window should not be empty after an RTT. However, I have failed to discover a definitely fair way for this...

- The HTTP/2 window change frame also requires an RTT to make the upstream refill the window.
- Therefore, to handle the high latency, the only way is to increase the window size.
- As we increase the HTTP/2 window size, we should also increase the local channel capacity to make them fair. Then how can we determine the capacity? 🥵 The first thing must be using the same "unit" for both flow control. As discussed in the doc, the "row count" might be the best solution.

I've tested it in https://github.com/risingwavelabs/risingwave/pull/6170.

rfcs/0015-permit-based-back-pressure.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Problems
+
+According to the recent benchmarks, we find a serious problem: since we bound the local exchange with the **chunk number** while the remote exchange with the **window size (bytes)**, which has different effects under different workloads, there can be significant imbalance between local and remote exchange.
+
+> [#6099 streaming: imbalance between local and remote inputs](https://github.com/risingwavelabs/risingwave/issues/6099)
+
+Take TPCH Q12 as an example:
+
+```jsx
```
TennyZhuang, line 27:
```suggestion
```plain
```

?

TennyZhuang (approved):
If the changes have been merged in the main repo, we can also merge the RFC PR.

PR #17, opened 2022-11-07
RFC: DML Design

st1page:
Preview: https://github.com/risingwavelabs/rfcs/blob/sts/DML/rfcs/0017-DML.md


rfcs/0017-DML.md
```diff
@@ -0,0 +1,55 @@
...
+We offer a syntax like that: `create table t1 (v1 int, v2 int) with (appendonly = true)`. Like some other OLAP systems, the "append only" indicate two things.
+
+1. user can not define primiary key or other unique constraints on that table
+2. user can not delete/update the table
+
+### Implementation details
+
+#### Materialize check
+
+The consistency and constraints check will be done in materialize executor. For every operation from upstream, the materialize executor will lookup its pk from storage and fix the changes if it is conflict with the storage's data. And because of the look up queries, a cache is needed.
```
BowenXiao1999, line 43: How to design the cache? 

If For every operation (INSERT\UPDATE\DELETE), the cost seems huge and hard to mitigate... Also I think it looks similar to Concurrent DELETE doc proposed by @fuyufjh a long time ago.

rfcs/0017-DML.md
```diff
@@ -0,0 +1,55 @@
...
+
+Currently we treat the DML as a kind of source. But now we need serve the data from both connector source and DML. So the `DML executor` is introduced to merge the DML changes stream into main stream. It will hold the `TableSource` as a RX to receive the operations from batch DML executors.
+
+#### Generate row_id in BatchInsert
+
+When user do an insert on a table without PK, we should help to generate the row_id for each records. Now we can only generate row_id in the SourceExecutor. But now, we might need to generate row id for both external data and DML insert's data. So the `BatchInsert` executor and `SourceExecutor` will share the `RowIdGenerator`.
+
+#### DML manager
+
+The `TableSourceManager` will rename as `DMLManager`. It will be indexed as `table_id` instead of `source_id` of the table source. The DML batch executors can get the `TableSource` for DML operation and the `Mutex<RowIdGenerator>` easily.
```
BowenXiao1999, line 55: DDL on MV will not go through SourceManager? 🤔

rfcs/0017-DML.md
```diff
@@ -0,0 +1,55 @@
...
+---
+
+# DML design
+
+## Motivation
+
+the [Unify the materialized source and table](https://github.com/risingwavelabs/rfcs/pull/4) will lead to 2 things
+
+- table's data are materialized in storage and we should check the consistency and constraints when materializing
+- table can both subscribe to some connectors and support `INSERT`, `DELETE,` and `UPDATE`, which meas it should serve two sources of data changes.
```
BowenXiao1999, line 15:
```suggestion
- table can both subscribe to some connectors and support `INSERT`, `DELETE,` and `UPDATE`, which means it should serve two sources of data changes.
```

rfcs/0017-DML.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Different objects's DML
+
+#### Source
+
+A source get append-only records from external source and `INSERT` statment and our system will generate a row_id for each record. User can not define PK and other constraints on that. Because all row id are generated internally, we do not need to check the pk's unique constraint.
+
+#### Table
+
+A table can get any operation from external source and `INSERT`, `UPDATE`, `DELETE` statement. The later operations can overwrite the previous ones. If user does not define a primary key, we will generate a row_id for each record. But because user can do modification on table, we must always do the check no matter if the pk is generated internally.
```
BowenXiao1999, line 29: A little unclear here. Insert into table is also insert into Source?

BowenXiao1999 (commented):
BTW, if we do not support primary key def on table, we should forbidden this behaviour (?) and report to user (Currently seems we do not do that).

rfcs/0017-DML.md
st1page: Yes, it is that doc mentions. The check is to ensure the external data is correct. And we should do the check to ensure the data is ok and do not lead to panic of other component in our system

rfcs/0017-DML.md
st1page: :thinking:

rfcs/0017-DML.md
st1page: After we adopt the design https://github.com/risingwavelabs/rfcs/pull/4 we will not have the `TableSource` because the table can be also get data from the external connector.

st1page:
DML

rfcs/0017-DML.md
fuyufjh: Yes.

The core difference here is that table itself is _a source of truth_ so we must allow people to do anything on it, including synchronize external changes on it or manually update/delete/insert into it.

rfcs/0017-DML.md
```diff
@@ -0,0 +1,57 @@
...
+
+#### Append Only Table(undocumented)
+
+To simplify the definitions in our system, that is a kind of undocumented feature and we do not encourage user do that. It is just for test now.
```
fuyufjh, line 33:
```suggestion
To simplify the definitions in our system, there is a kind of undocumented feature and we do not encourage user do that. It is just for test now.
```

rfcs/0017-DML.md
```diff
@@ -0,0 +1,57 @@
...
+
+The consistency and constraints check will be done in materialize executor. For every operation from upstream, the materialize executor will lookup its pk from storage and fix the changes if it is conflict with the storage's data. And because of the look up queries, a cache is needed.
+
+#### DML executor
+
+Currently we treat the DML as a kind of source. But now we need serve the data from both connector source and DML. So the `DML executor` is introduced to merge the DML changes stream into main stream. It will hold the `TableSource` as a RX to receive the operations from batch DML executors.
+
+#### Generate row_id in BatchInsert
+
+When user do an insert on a table without PK, we should help to generate the row_id for each records. Now we can only generate row_id in the SourceExecutor. But now, we might need to generate row id for both external data and DML insert's data. So the `BatchInsert` executor and `SourceExecutor` will share the `RowIdGenerator`.
```
fuyufjh, line 51: > So the `BatchInsert` executor and `SourceExecutor` will share the `RowIdGenerator`

Please allow me to clarify that I don't have any insistence on this part. Our current approach is to let the `BatchInsert` insert rows with row_id `NULL`, which is acceptable to me. I just hope to keep it simple as possible, so let's see which approach is better when writing the code. 😋

fuyufjh:
LGTM.

fuyufjh:
Besides, We can move `Exchange` before `DMLExec` if we can

1. Schedule batch DML executors according to the PKs
2. Generate RowIDs of a certain vnode

fuyufjh:
By the way, in the future we may need to optionally support persisting users' writes somewhere, to ensure no `insert`s will be lost even on failure recovery. To achieve that, we will refactor the channel between batch DML executor and streaming `DMLExec` to a MQ with high availablity like Kafka or SQS, and it seems to be another reason for using a single channel for insert/delete/update instead of two.

st1page:
image

xx01cyx (commented):
The first query description in the figure should be "no user defined pk". Rest LGTM.

st1page:
dml_new drawio

xx01cyx (approved):

PR #18, opened 2022-11-08
RFC: Temporal Filter

fuyufjh:
Preview: https://github.com/risingwavelabs/rfcs/blob/eric/0018-temporal-filter/rfcs/0018-temporal-filter.md

Migrated from https://singularity-data.quip.com/a6XXAQVa6b0N/RFC-Temporal-Filter

### DIFF
--- a/rfcs/0018-temporal-filter.md
+++ b/rfcs/0018-temporal-filter.md
@@ -1,0 +1,115 @@
+---
+feature: temporal_filter
+authors:
+  - "Eric Fu"
+start_date: "2022/11/08"
+---
+
+# RFC: Temporal Filter
+
+## Background
+
+A temporal filter is a filtering condition on some particular time column according to the current time, usually used to **represent a sliding window** with a flexible start or end. For example,
+
+* `time_col >= now() - INTERVAL '1 hour'` 
+    — a sliding window of recent 1 hour
+* `time_col BETWEEN now() - INTERVAL '2 hour' AND NOW() - INTERVAL '1 hour'` 
+    — a sliding window of 1 hour before 1 hour
+* `time_col >= date_trunc('day', now())` 
+    — a sliding window of today (from 0:00)
+
+On the contrary, below are ***NOT*** temporal filters
+
+* `time_col >= '2022-10-01 00:00:00'` 
+    — an ordinary static filtering condition
+* `to_char(time_col, ’yyyymmdd') = to_char(now(), 'yyyymmdd')
+    `— the condition is not applied on datetime/timestamp data type
+
+Besides the semantics itself, a temporal filter is also useful to **bound the size of states for an arbitrary query**. For example,
+as mentioned in [State Cleaning in Flink](https://singularity-data.quip.com/Z4KPAzFlFKz8), a query without any proper temporal conditions naturally involve all historical data, thus the size of states will grow infinitely. To solve this, one possible solution is to put a temporal filter there:
+
+* `SELECT customer_id, amount FROM orders GROUP BY customer_id
+    `— the size of states is unbounded
+* `SELECT customer_id, amount FROM orders GROUP BY customer_id
+    WHERE order_date > NOW() - INTERVAL '7 DAYS'
+    `— the size of states is bounded
+
+
+You may also be interested in [the implementation of temporal filters in Materialize](https://materialize.com/blog/temporal-filters/#appendex-implementation). It’s quite clever, but unluckily, not applicable for RisingWave. I would strongly recommend taking a glance at that blog article, and you will actually find some connectivity between these 2 approaches - at least equivalent time & space compelexity.
+
+
+## Design
+
+### Lower-bound temporal conditions
+
+A lower-bound temporal condition let the incoming rows live for some bounded time, and will delete them at some time (mostly) in some future time point.
+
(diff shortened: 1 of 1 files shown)


soundOfDestiny:
Seems that we will add a new field `time_now` in our barrier.
Do we want to collect one result per barrier, or collect multiple results per barrier?
- If we collect one result per barrier, we must ensure that the `time_now` is greater than watermark, or the result will be meaningless.
- If we collect multiple results per barrier, should we produce `2N` results per barrier (given that `N` is num of different value in time column recently) or produce `M` results per barrier (given that `M` is num of barriers recently)?

jon-chuang:
Since we are doing a range delete for `TemporalFilter`, we should definitely use the delete range   interface (https://github.com/risingwavelabs/risingwave/issues/5978) for better performance.

jon-chuang:
@soundOfDestiny , am I wrong, or is `time_now` equivalent to the watermark? It seems we should issue `time_now` from meta, and ensure that all of the barriers of same epoch have the same `time_now`?

soundOfDestiny:
> @soundOfDestiny , am I wrong, or is `time_now` equivalent to the watermark? It seems we should issue `time_now` from meta, and ensure that all of the barriers of same epoch have the same `time_now`?

yes, I think that we should issue `time_now` from meta, and ensure that this `time_now` is greater than any watermark in source.

soundOfDestiny:
> @soundOfDestiny , am I wrong, or is `time_now` equivalent to the watermark? It seems we should issue `time_now` from meta, and ensure that all of the barriers of same epoch have the same `time_now`?

yes, I think that we should issue `time_now` from meta, and ensure that this time_now is less than latest watermark in each shard of source.

jon-chuang:
Will we also include an operator `insertion_time(t)` where t is table name?

So, for instance, 
```SQL
CREATE MATERIALIZED VIEW v AS SELECT * FROM t1 NATURAL JOIN t2 WHERE insertion_time(t1) > now() - '1h' AND insertion_time(t2) > now() - '15m'
```

jon-chuang:
> time_now is less than latest watermark in each shard of source.

By watermark, you are referring to compaction watermark? Or watermark used in watermark filter?

soundOfDestiny:
> > time_now is less than latest watermark in each shard of source.
> 
> By watermark, you are referring to compaction watermark? Or watermark used in watermark filter?

watermark used in watermark filter

fuyufjh:
> Seems that we will add a new field time_now in our barrier.

It's Optional. The epoch (int64) is designed to carry physical time with higher bits in it.

> Do we want to collect one result per barrier, or collect multiple results per barrier?

Do you mean when to emit results? In my mind, if an event arrives and satisfies the time condition, it will be passed to downstream immediately. But the shifting of time is driven by barrier, so when the barriers some, some `DELETE` events may be emitted to delete the events that not satisfied the temporal filter condition any more.

fuyufjh:
> Note that in this schemem, state cleaning and emitting `delete` events are independent. State cleanning is driven by `Watermark` message, that is, the events below the watermark will be removed from state. Emitting `delete` (or `insert`) events is driven by `Barrier` message, because the epoch in barriers will advance the current time (`now()`) and then trigger new `delete` or `insert` events to downsteam.

soundOfDestiny:
> > Seems that we will add a new field time_now in our barrier.
> 
> It's Optional. The epoch (int64) is designed to carry physical time with higher bits in it.
> 
> > Do we want to collect one result per barrier, or collect multiple results per barrier?
> 
> Do you mean when to emit results? In my mind, if an event arrives and satisfies the time condition, it will be passed to downstream immediately. But the shifting of time is driven by barrier, so when the barriers some, some `DELETE` events may be emitted to delete the events that not satisfied the temporal filter condition any more.

I'm not sure the result at the barrier point it meaningful because it does not contain all data before `time_now`.

jon-chuang:
> Either introducing a new TemporalFilter operator or reusing DynamicFilter looks good to me. Let's leave the choice to the implementer.

I think regardless, I'd prefer the use of now() emitting watermark, as that means that something downstream of the Dynamic/TemporalFilter can also clean its state based on the filter's timestamp field.

soundOfDestiny:
> > Either introducing a new TemporalFilter operator or reusing DynamicFilter looks good to me. Let's leave the choice to the implementer.
> 
> I think regardless, I'd prefer the use of now() emitting watermark, as that means that something downstream of the Dynamic/TemporalFilter can also clean its state based on the filter's timestamp field.

yes, it will.

fuyufjh:
It seems to be completed, right? Let's approve and merge it. cc. @soundOfDestiny @jon-chuang

jon-chuang:
@fuyufjh We have yet to add e2e test and also missing forward of Watermark on Project operator (used for expressions of form `now() +- interval`) https://github.com/risingwavelabs/risingwave/issues/6790

soundOfDestiny (approved):
sadly introducing watermark will still meet `todo!` panic in our executors

soundOfDestiny:
Datatype of `now()` in batch and in streaming are now both implemented as `Timestamptz`.
`now()` in streaming will return time of the barrier epoch.
`now()` in batch will return time of the latest epoch that the binder saw.

PR #19, opened 2022-11-09
RFC: S3 Source

tabVersion:
preview: https://github.com/risingwavelabs/rfcs/blob/0cff6144ffdfdfe91ea3d1217d67f0b09b69f7ea/rfcs/0019-s3%20source.md

### DIFF
--- a/rfcs/0019-s3 source.md
+++ b/rfcs/0019-s3 source.md
@@ -18,13 +18,13 @@ Support load data from Amazon S3 and other Simple Storage Services.
 More and more scenarios require us to support data synchronization from S3, which challenges our existing abstractions.
 The main problem in the previous design is that a source could be read many times by multiple materialized views but the SQS message can only be consumed once, which leads to data inconsistency.
 
-In the source abstraction, one source can only have one enumerator to discover the upstream changes. However, the executors in different fragments have different consuming process and different state table.
-It is a trivial idea to discover the changes by every executor and the enumerator only confines the concurrency.
+First of all, we believe that s3 source is append only. This is consistent with the semantics of s3, which does not allow a file to be modified, but only to delete the old one and add the new one. So s3 source will re-read the full amount of updated files.
 
-The source executor will start a thread when starting a S3 connector and list all objects in the given bucket.
-After a regex match, the object names will be passed to reader by a channel. 
+In the source abstraction, one source can only have one enumerator to discover the upstream changes. Although we cannot implement multiple consumptions on an SQS, we can have the enumerator cache the queue results.
 
-- [ ] The channel can break the abstraction of source, we may consider a new file-system source executor apart from the original one.
+At the first start, the enumerator first `ls` the object in the bucket and stores it in a hashset for persistence, at which point we consider the source creation complete. Then it starts to consume SQS and maintains a list of all objects that need to be consumed by SQS.
+
+When the source executor is started, the elements in the hashset are assigned to the specific executor by a consistent hash, and the new added objects are distributed by config change. One of the optimization points here is to transfer only the changed paths, which reduces memory consumption.
 
 ### Scale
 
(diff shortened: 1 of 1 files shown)


rfcs/0019-s3 source.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Design
+
+More and more scenarios require us to support data synchronization from S3, which challenges our existing abstractions.
+The main problem in the previous design is that a source could be read many times by multiple materialized views but the SQS message can only be consumed once, which leads to data inconsistency.
+
+In the source abstraction, one source can only have one enumerator to discover the upstream changes. However, the executors in different fragments have different consuming process and different state table.
+It is a trivial idea to discover the changes by every executor and the enumerator only confines the concurrency.
+
+The source executor will start a thread when starting a S3 connector and list all objects in the given bucket.
```
liurenjie1024, line 24: Does this mean that every executor will scan whole bucket periodically?

rfcs/0019-s3 source.md
```diff
@@ -0,0 +1,59 @@
...
+In this design the enumerator always provides M hashslots and assigns these slots to N nodes. 
+For example, the enumerator provides 256 (0~255) slots and there are three executors, each partition will look like:
+
+```plain
+S3Partition{ "min_slot": 0, "max_slot": 85 }
+S3Partition{ "min_slot": 86, "max_slot": 171 }
+S3Partition{ "min_slot": 171, "max_slot": 255 }
+```
+
+Each executor get the object names and use a hash function to filter out the names meet the restriction.
```
liurenjie1024, line 40: Is it possible that user delete an object, and then put an object with same name? In this case we may miss the new one.

liurenjie1024:
Maybe we can learn from delta lake and use a self maintained log to store changes from SQS, this way different consumer with same source can reuse this log. We can also assign an id to each found object so that we can remove already consumed object.

rfcs/0019-s3 source.md
```diff
@@ -0,0 +1,59 @@
...
+
+Each executor get the object names and use a hash function to filter out the names meet the restriction. 
+We can use consistent hash here for load balancing.
+
+The scale in and scale out operations are intuitive, just adjust the `min_slot` and `max_slot` for each partition.
+
+### Failover
+
+As we only sustain `(object_name, latest_offset)` in the state store, it is unacceptable to download all objects and check whether the reader has reached the end.
+The best practice here is that if the object name is in the state table, it is considered have been read, where we bring a strict rule that **every object must be read completely within one epoch**.
```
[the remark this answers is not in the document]
tabVersion: > This may slowdown epoch is user have a large object.

allowing splitting one object into multiple epochs may increase the complexity significantly. 

> Also what's the strategy of cleaning consumed object?

we are not removing consumed objects. If an object name in the state table expires, we will treat it as a new object.

rfcs/0019-s3 source.md
tabVersion: yes, we let the executor maintain its own state, just as flink does.

rfcs/0019-s3 source.md
tabVersion: yes, we will miss the change. We are suffering from ecosystem here, AWS provides SDK in other languages which can list objects filtered by timestamp but in rust it cannot.

tabVersion:
> Maybe we can learn from delta lake and use a self maintained log to store changes from SQS, this way different consumer with same source can reuse this log.

I will look into this solution but our enumerator is stateless and this state can be extensive and it is not proper to keep it in etcd. @yezizp2012 can we let meta directly interact with storage?

> We can also assign an id to each found object so that we can remove already consumed object.

I think a LSM tree like structure can do this

yezizp2012:
> > Maybe we can learn from delta lake and use a self maintained log to store changes from SQS, this way different consumer with same source can reuse this log.
> 
> I will look into this solution but our enumerator is stateless and this state can be extensive and it is not proper to keep it in etcd. @yezizp2012 can we let meta directly interact with storage?

This's much more like create materialized source for s3 source by default. One solution might works for it:
1. we can create materialized source for s3 source: (every record in s3 object with be added with an additional column 'object name', if we want to not only handle append-only s3 source but also the scenario that object could be removed mentioned by renjie:
> Is it possible that user delete an object, and then put an object with same name? In this case we may miss the new one.
2. Then all materialized view upon S3 source will be MoM, the enumerator will consume the logs from SQS, and generate chunks by SQS message types:
- create/append: insert records.
- delete: remove all records with specific object name.

I'm not familiar with SQS, correct me if I'm wrong.

liurenjie1024:
@yezizp2012 's idea sounds good to me. This way we can avoid 
1. Too frequent scan of bucket.
2. Missing renamed object.

liurenjie1024:
Also if we generate an auto incremental id for each object version, we can safely remove consumed object version in state store.

rfcs/0019-s3 source.md
skyzh: I think you can do that with aws-s3-sdk. It should be the same as SDK in other languages. They are generated from some kind of template.

skyzh:
> delete: remove all records with specific object name

1. We can assume s3 source is append-only and delete is no-op?
2. It doesn't seem possible to do all the deletions within one epoch.

tabVersion:
> > delete: remove all records with specific object name
> 
> 1. We can assume s3 source is append-only and delete is no-op?

agree. I prefer s3 source is append-only

---

> Maybe we can learn from delta lake and use a self maintained log to store changes from SQS, this way different consumer with same source can reuse this log. We can also assign an id to each found object so that we can remove already consumed object.

My concern is this solution may lead to frequent config change, which can harm the source reader's efficiency.
Besides, meta is responsible for maintaining every actor's split assignment and pushing it down in failover scenes. It can be tricky to design what the connector persists. 

---

> Then all materialized view upon S3 source will be MoM, the enumerator will consume the logs from SQS, and generate chunks by SQS message types:
> * create/append: insert records.
> * delete: remove all records with specific object name.

I don't know if it is a common practice to ask users to provide both s3 and sqs. Do we need to provide a solution for users who do not provide sqs. @neverchanje

StrikeW:
Could you provide some use cases of S3 sources? What does the schema of the S3 source look like? And what computations that user can do on S3 objects？

fuyufjh:
Whether to depend on SQS is a crucial problem here. 

As mentioned by @waruto210, [Flink's FileSystem connector](https://nightlies.apache.org/flink/flink-docs-master/docs/connectors/table/filesystem/) simply scan the file in a fixed interval. The might be the most general cases, and most users don't even know that S3 can be configured to emit events into SQS.

tabVersion:
Will propose a new S3 Source design. Close this one

PR #20, opened 2022-11-10
RFC: Batch query on source.

liurenjie1024:


rfcs/batch_query_on_source.md
```diff
@@ -0,0 +1,70 @@
...
+ROW FORMAT JSON;
+
+select * from source_abc limit 10;
+```
+
+In general, we need to add source plan node for batch plan and source scan executor for batch query engine. The parallelism of source scan executor is determined by the number of source splits, which can be fetched from sources.
+
+### Extra columns
+
+By default, we allow user to scan data using earliest or latest offset. This default option works well in streaming, but maybe not suitable for batch execution. For example, user may just want to scan kafka data in latest 10 minutes, or lastest 10000 messages. Or when user has already partitioned data, they may want to query only some partition. To fufill these requirements, we can expose message meta as extra columns in schema. Let's use kafka as an example, we can attach following extra columns to each source from kafka:
```
TennyZhuang, line 48: Flink allows users to extract such fields by the `METADATA FROM` clause.

rfcs/batch_query_on_source.md
```diff
@@ -0,0 +1,70 @@
...
+
+| Column name                | Type      | Comment                                                                                   |
+|----------------------------|-----------|-------------------------------------------------------------------------------------------|
+| _rw_kafka_partition        | int       | Kafka partition this message from.                                                        |
+| _rw_kafka_partition_offset | bigint    | Kafka offset in partition. 0 means starts from earliest, and -1 means starts from latest. |
+| _rw_kafka_timestamp        | timestamp | Message timestamp                                                                         |
+
+With these extra columns, we can implement several optimizations:
+
+1. To query messages from some time, we can write `_rw_kafka_timestamp > '2022-11-10 10:00:00'`. We can use kafka's seek time offset api to determine smallest offset later than `'2022-11-10 10:00:00'`.
```
fuyufjh, line 58: So if user doesn't specify the condition in the `where` condition, all available events will be selected regardless of `scan.startup.mode`, right?

rfcs/batch_query_on_source.md
liurenjie1024: Yes, they are similar, both are virtual columns.

rfcs/batch_query_on_source.md
```diff
@@ -0,0 +1,70 @@
...
+
+In general, we need to add source plan node for batch plan and source scan executor for batch query engine. The parallelism of source scan executor is determined by the number of source splits, which can be fetched from sources.
+
+### Extra columns
+
+By default, we allow user to scan data using earliest or latest offset. This default option works well in streaming, but maybe not suitable for batch execution. For example, user may just want to scan kafka data in latest 10 minutes, or lastest 10000 messages. Or when user has already partitioned data, they may want to query only some partition. To fufill these requirements, we can expose message meta as extra columns in schema. Let's use kafka as an example, we can attach following extra columns to each source from kafka:
+
+| Column name                | Type      | Comment                                                                                   |
+|----------------------------|-----------|-------------------------------------------------------------------------------------------|
+| _rw_kafka_partition        | int       | Kafka partition this message from.                                                        |
```
neverchanje, line 52: What if the source is not Kafka?

liurenjie1024: I'm using kafka as an example here. Other source may use similar strategy to help prunning data to scan.

rfcs/batch_query_on_source.md
```diff
@@ -0,0 +1,70 @@
...
+authors:
+  - "Renjie Liu"
+start_date: "2022/11/10"
+---
+
+# Batch Query on Source
+
+## Summary
+
+Introduce execute batch query on sources like kafka/s3.
```
TennyZhuang, line 12:
```suggestion
Support executing the batch queries on sources like Kafka/S3.
```

rfcs/batch_query_on_source.md
neverchanje: Ok so for pulsar will there be an `_rw_pulsar_timestamp` or should we rename this var to some more general?

liurenjie1024: Good question, I think both are ok to me. What's important for user is to have clear document for meaning and use example.

rfcs/batch_query_on_source.md
fuyufjh: Alternatively I think we can also ignore `scan.startup.mode` and always read from earlest, unless specified by `where _rw_kafka_timestamp > ...`

liurenjie1024: Anyway, I think a better solution is to separate connector/source definition to make things clear and explicit.

fuyufjh: Yes. In theory the clearest way is to differentiate `create external table` and `create source`, while the problem is that we don't want users to define it twice.

rfcs/batch_query_on_source.md
```diff
@@ -0,0 +1,70 @@
...
+
+| Column name                | Type      | Comment                                                                                   |
+|----------------------------|-----------|-------------------------------------------------------------------------------------------|
+| _rw_kafka_partition        | int       | Kafka partition this message from.                                                        |
+| _rw_kafka_partition_offset | bigint    | Kafka offset in partition. 0 means starts from earliest, and -1 means starts from latest. |
+| _rw_kafka_timestamp        | timestamp | Message timestamp                                                                         |
```
xiangjinwu, line 54: This is better modeled as `timestamp with time zone` rather than `timestamp`.
* Internally, `timestamp with time zone` stores the (micro)seconds since epoch in UTC while `timestamp` stores a year-month-day-hour-minute-second tuple
* On usage, `timestamp with time zone` allows the user to write `_rw_kafka_timestamp > '2022-11-10 10:00:00Z'` or `_rw_kafka_timestamp > '2022-11-10 02:00:00-08:00'`, but `timestamp` requires a PST user to write `_rw_kafka_timestamp > '2022-11-10 10:00:00'` when the desktop clock is saying `2022-11-10 02:00:00`.

liurenjie1024: Good point!

rfcs/batch_query_on_source.md
```diff
@@ -0,0 +1,70 @@
...
+
+With these extra columns, we can implement several optimizations:
+
+1. To query messages from some time, we can write `_rw_kafka_timestamp > '2022-11-10 10:00:00'`. We can use kafka's seek time offset api to determine smallest offset later than `'2022-11-10 10:00:00'`.
+2. To limit scan by range of offsets, we can use `_rw_kafka_partition_offset > 1000`.  If the offset range is postive number, it's counted from earliest. We can use `_rw_kafka_partition_offset > -1000` to express latest 1000 message in each partition.
```
tabVersion, line 59: I doubt whether kafka provides indices by a negative number.

liurenjie1024: We will first only implement timestamp based offsets.

rfcs/batch_query_on_source.md
```diff
@@ -0,0 +1,70 @@
...
+3. To limit partition to scan, we can use `_rw_kafka_partition = 0` to scan data from partition 0 only.
+
+Above queries use kafka as an example, we can have similar stategies for other sources(pulsar, kinesis, etc).
```
fuyufjh, line 62: By the way, even if these columns are used for filtering majorly, we also need to implement on the output side to make it compliant with SQL's semantics i.e. `select _rw_kafka_timestamp, _rw_kafka_partition_offset, _rw_kafka_partition, * from my_source` should give correct results.

xxchan:
Tracking issue https://github.com/risingwavelabs/risingwave/issues/6358 is closed. Merge? (Anything needs to be updated?)

liurenjie1024:
> Tracking issue https://github.com/risingwavelabs/risingwave/issues/6358 is closed. Merge? (Anything needs to be updated?)

LGTM

PR #21, opened 2022-11-15
add RFC: Backup and Restore

zwang28:
[Preview](https://github.com/risingwavelabs/rfcs/blob/d923c366328c258ef1aaf4ef7cf0f4419ec5783b/rfcs/0000-backup-and-restore.md)

backup drawio

### DIFF
--- a/rfcs/0000-backup-and-restore.md
+++ b/rfcs/0000-backup-and-restore.md
@@ -1,0 +1,42 @@
+---
+feature: backup-and-restore
+authors:
+  - "zwang28"
+start_date: "2022/11/14"
+---
+
+# Backup and Restore
+
+## Motivation
+Support backup and restore to protect against data loss, due to user error, software error, disaster scenario, etc.
+
+## Design
+
+A RisingWave cluster's persistent data includes metadata stored in meta store, and SSTable stored in object store. We back up these types of data differently:
+- For metadata, we back up a metadata snapshot. See [Metadata Snapshot](#metadata-snapshot).
+- For SSTable, we decide **not to replicate object store data** during backup. Instead, we only back up a SSTable manifest, which references to the same SSTable as the running cluster. See [SSTable Maintenance](#sstable-maintenance).
+- In addition, to support point-in-time recovery(PITR), we also back up metadata changes. See [Point-in-Time Recovery](#point-in-time-recovery).
+
+### Metadata Snapshot
+We leverage meta store's snapshot read to get a consistent cut of all interested metadata. Specifically with etcd meta store, we read using the same etcd revision during backing up.
+
+### SSTable Maintenance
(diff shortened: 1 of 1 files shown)


rfcs/0000-backup-and-restore.md
```diff
@@ -0,0 +1,44 @@
...
+- For SSTable, we decide **not to replicate object store data** during backup. Instead, we only back up a SSTable manifest, which references to the same SSTable as the running cluster. See [SSTable Maintenance](#sstable-maintenance).
+- In addition, to support point-in-time recovery(PITR), we also back up metadata changes. See [Point-in-Time Recovery](#point-in-time-recovery).
+
+![](https://user-images.githubusercontent.com/70626450/201866284-923db172-89ff-423d-95e0-0e93bcde23da.png)
+
+### Metadata Snapshot
+We leverage meta store's snapshot read to get a consistent cut of all interested metadata. Specifically with etcd meta store, we read using the same etcd revision during backing up.
+
+### SSTable Maintenance
+Now that SSTables are shared between the running cluster and backups, a SSTable shouldn't be deleted until it's not referenced by any backup's SSTable manifest. Specifically, a running cluster is aware of all available backups. When trying to delete stale SSTables, it filters out candidate SSTables by checking all backup's SSTable manifest.
```
chenzl25, line 26: I have some concerns about whether multiple clusters running with the same set of SSTables would corrupt the SSTables? e.g. different compactor and different vacuum from multiple clusters.

zwang28: We shouldn't allow "multiple clusters running with the same set of SSTables". That means we must ensure the original cluster is entirely stopped before restoring a new one. Indeed it sounds like some operational risk.

rfcs/0000-backup-and-restore.md
```diff
@@ -0,0 +1,44 @@
...
+# Backup and Restore
+
+## Motivation
+Support backup and restore to protect against data loss, due to user error, software error, disaster scenario, etc.
+
+## Design
+
+A RisingWave cluster's persistent data includes metadata stored in meta store, and SSTable stored in object store. We back up these types of data differently:
+- For metadata, we back up a metadata snapshot. See [Metadata Snapshot](#metadata-snapshot).
+- For SSTable, we decide **not to replicate object store data** during backup. Instead, we only back up a SSTable manifest, which references to the same SSTable as the running cluster. See [SSTable Maintenance](#sstable-maintenance).
```
Little-Wallace, line 17: What if users hope to backup data to other storage?

zwang28: I think we can optionally support incremental SSTable replication.

Li0k (approved):

zwang28:
I propose to prioritize snapshot backup and time travel query.
After that we can consider PITR.

PR #22, opened 2022-11-18
RFC: S3 Source with SQS

tabVersion:
preview https://github.com/risingwavelabs/rfcs/blob/tab/s3-design-2/rfcs/0022-s3-source.md


rfcs/0022-s3-source.md
```diff
@@ -0,0 +1,75 @@
...
+## Motivation
+
+Support load data from Amazon S3 and other Simple Storage Services.
+
+## Background
+
+The previous two versions RFCs are available as follows:
+
+* [Version 1](https://www.notion.so/risingwave-labs/Deprecated-RFC-S3-Source-Design-632aa78264b347f9a4a37fd89a91d785): presents the basic abstraction of treating a file as a partition, but here we ignore the fact that a source can be relied on by multiple MVs, while sqs can only be consumed once. This results in inconsistent data across multiple MVs.
+* [Version 2](https://github.com/risingwavelabs/rfcs/pull/19): want the executors to list s3 buckets independently and maintain their own state. This solution is not accepted because listing the bucket periodically is too expensive ([price policy](https://aws.amazon.com/s3/pricing/)) and we want to use SQS to describe the incremental data.
```
fuyufjh, line 19: The combinition of S3 + SQS is very rare in my (limited) experience. I have never seen any system supported for it. Even Airbyte only supports the `ls` approach (https://airbytehq.github.io/integrations/sources/s3/).

My concern is that in most cases, users may not have full control over his S3 buckets, thus, he is unable or unwilling to turn on SQS for RisingWave. I suppose this could be an optional feature and more than 2/3 users don't use it.

Personally I recommend to focus on S3 **without** SQS at this time.

fuyufjh: cc. @neverchanje @tabVersion

fuyufjh: If we agree to give up SQS support, the design can also be simplified a lot:

- The `Chain` operator between `List` and `Fetch` is not needed anymore
- The `List` stage can be completely stateless

Actually, I guess the whole design can be refactored...

fuyufjh:
If we agree to give up SQS support, I think we can fall back to the "classical" solution - the way that we intend when designing the source module.

It is worth mentioning that our design was basically borrowed from Flink: 

- [Data Sources | Apache Flink](https://nightlies.apache.org/flink/flink-docs-master/docs/dev/datastream/sources/)
- [FileSystem Connector | Apache Flink](https://nightlies.apache.org/flink/flink-docs-master/docs/connectors/datastream/filesystem/)

As you can see, the concepts such as `Split` and `SplitEnumerator` are exactly the same as ours.

Note that `SplitEnumerator` has its own state, which is a list of discovered splits as well as their assigned executors. For Flink, it's part of the checkpoints; while for us, it's stored in Meta Service. 

Specifically,

- For Kafka source, the state is a mapping from the Kafka partition to their assigned executors
- For S3/File source, the state is a mapping from object/file name to thier assigned executors

Well, that's all. All things work! 👻

1. The `SplitEnumerator` on the Meta Service list the objects/files
2. Once new files are discovered, `SplitEnumerator` informs the `SourceExecutor`s with a `Mutation` barrier
3. The `SourceExecutor` keeps the read offset as their state

You may complain that the file list can be very huge. That's actually not a big problem.

- First, for a very large mutation (like initialization), we can do it in small batches. Actually, S3's List API can return a maximum of 1000 files at one time. 
- For other mutations than the initialization one, we can only send diff instead of the whole list. Very natural idea because both sides are stateful.

Looking forward to your comments.

PR #23, opened 2022-11-21
RFC: Minimize Height Join Ordering

chenzl25:
image
image
image

### DIFF
--- a/rfcs/0000-minimize-height-join-ordering.md
+++ b/rfcs/0000-minimize-height-join-ordering.md
@@ -1,0 +1,87 @@
+---
+feature: minimize_height_join_ordering
+authors:
+  - "Dylan Chen"
+start_date: "2022/11/21"
+---
+
+# Minimize Height Join Ordering
+
+## Motivation
+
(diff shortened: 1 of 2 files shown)


skyzh:
I wonder if we reduce the height of join, is it possible that we might need to materialize more intermediate data of join? (Not sure if there is mathematics way to compute this.)

chenzl25:
> I wonder if we reduce the height of join, is it possible that we might need to materialize more intermediate data of join? (Not sure if there is mathematics way to compute this.)

Another cost function we can use is the intermediate result of all the operators just like what you mentioned before.
If we know the statistics in advance, we can calculate the intermediate result of a join output as `|left| * |right| * selectivity`. 
However, without any statistics it is hard to tell which join order will produce the minimal intermediate result. Reducing the height of the join tree seems like it can reduce the barrier latency definitely, so I think we should achieve something which is sure to have benefits rather than something we can't predict.

xxchan (approved):
Completed at https://github.com/risingwavelabs/risingwave/issues/8316

PR #24, opened 2022-11-21
RFC: materialize the result to handle non-deterministic functions

st1page:


rfcs/0024-nondeterministic-functions.md
```diff
@@ -0,0 +1,60 @@
...
+The `StreamMaterializeProject` will materialize the first columns of its input which is the `random()` function's result. When we have a serial of changes `Insert(v = 10, pk = 1), Update((v = 10, pk = 1) -> (v = 20, pk = 1)), Delete(v = 20, pk = 1)` the `StreamMaterializeProject` will handle it as following:
+
+1. receive `Insert(v = 10, pk = 1)`
+    - comput the result `(rd = 111, v = 20, pk = 1)`
+    - insert into state table `(pk = 1, rd = 111)`
+    - output `Insert(rd = 111, v = 20, pk = 1)`
+
+2. receive `Update((v = 10, pk = 1) -> (v = 20, pk = 1))`
+    - compute the old value's project result `(rd = 55555, v = 20, pk = 1)`
+    - get the old values `(pk = 1, rd = 111)` from state table modify the result old value to `(rd = 111, v = 10, pk = 1)`
```
skyzh, line 42:
```suggestion
    - get the old values `(pk = 1, rd = 111)` from state table modify the result old value to `(rd = 111, pk = 1)`
```

rfcs/0024-nondeterministic-functions.md
```diff
@@ -0,0 +1,60 @@
...
+    - compute the new value `(rd = 444, v = 30, pk = 1)`
+    - insert into state table `(pk = 1, rd = 444)`
+    - output `Update((rd = 111, v = 20, pk = 1) -> (rd = 444,v = 40, pk = 1))`
```
skyzh, line 45: if users execute `UPDATE SET v = 40`, is it expected that `rd` will also change?

rfcs/0024-nondeterministic-functions.md
```diff
@@ -0,0 +1,60 @@
...
+
+We will introduce a `MaterializeProject`, it will materialize some **partial** columns of result with the stream key of the input as the primiary key. When an `Insert` operation comes, it will compute the result and materialize some columns. when `Update` or `Delete` comes, it will lookup its state to replace the old value of the operation.
+For example, if we have a materialized view with SQL `select random() as rd, v*2 as vv, pk from t` and the pk is the primary key of the table t. The stream processing plan will be like:
+
+```plain
+StreamMaterialize(columns:[rd, vv, pk])
+  StreamMaterializeProject(
+      exprs: [Random(), InputRef(0) * 2, InputRef(1)],
+      materialized_cols: [0],
+      state_table: [pk, rd]
```
skyzh, line 28: need to emphasis state table key = pk

rfcs/0024-nondeterministic-functions.md
```diff
@@ -0,0 +1,60 @@
...
+    )
+    StreamScan(fields:[v, pk])
+```
+
+The `StreamMaterializeProject` will materialize the first columns of its input which is the `random()` function's result. When we have a serial of changes `Insert(v = 10, pk = 1), Update((v = 10, pk = 1) -> (v = 20, pk = 1)), Delete(v = 20, pk = 1)` the `StreamMaterializeProject` will handle it as following:
```
skyzh, line 33: The `StreamMaterializeProject` will materialize all output of projection with non-deterministic result.

rfcs/0024-nondeterministic-functions.md
st1page: I mean that the modification is on the old value's project result `(rd = 55555, v = 20, pk = 1)` I will express it more detailed

rfcs/0024-nondeterministic-functions.md
xxchan, line 1: FWIW (not read yet), flink has a [`TRY_RESOLVE`](https://nightlies.apache.org/flink/flink-docs-master/docs/dev/table/concepts/determinism/#3-determinism-in-streaming-processing) and [`PLAN_ADVICE`](https://nightlies.apache.org/flink/flink-docs-release-1.17/docs/dev/table/sql/explain/#explaindetails) to mitigate non-determinism. Are you aware of that?

xxchan: Oh, it seems `TRY_RESOLVE` is just adding materialization.

rfcs/0024-nondeterministic-functions.md
```diff
@@ -0,0 +1,60 @@
...
+
+# nondeterministic-functions
+
+## Summary
+
+We introduce the `MaterializeProject` stream executor to handle non-deterministic function on updatable stream.
+
+## Motivation
+
+In the discussion about the udf, @skyzh raise the issue about the non-deterministic function on updatable stream <https://github.com/risingwavelabs/rfcs/pull/12#issuecomment-1305105614>. This issue also appears on the built-in functions such as `random(),` `proc_time()` or `now()`.
```
xxchan, line 16: For UDF, how can we decide the determinism? Specified by users? 🤔

xxchan: BTW, from @jon-chuang's demo idea, is "caching expensive UDF" another valid motivation? 🤔

st1page: Can we trust it? We currently maintain very high consistency in our system and a wrong old value could produce unsolvable panic... We even do not totally trust the CDC source now... :thinking:

xxchan: Then always materializing sounds good to me 🤪

jon-chuang: Caching is an application-level concept... the user needs to be sure they are OK with stale values.

So yes, updateable stream is root cause.

jon-chuang: Agreed on always materializing

rfcs/0024-nondeterministic-functions.md
```diff
@@ -0,0 +1,60 @@
...
+authors:
+  - "st1page"
+start_date: "2022/11/21"
+---
+
+# nondeterministic-functions
+
+## Summary
+
+We introduce the `MaterializeProject` stream executor to handle non-deterministic function on updatable stream.
```
jon-chuang, line 12: Prefer that Project executor simply has optional state table with columns only corresponding to either UDF or non-deterministic function

jon-chuang:
Any interaction of UDF with watermark?

st1page:
> Any interaction of UDF with watermark?

I think we should give user way to config on the UDF such as 
- is deterministic, false by default
- is monotonically increasing, false by default

rfcs/0024-nondeterministic-functions.md
st1page: LGTM

BugenZhao:
This proposal inspires me that, if we're going to fix a bug for our built-in expression in a version update, it also breaks the determinism, even if the expression itself is mathematically pure. 😄 If we're for the sake of serious consistency and determinism, we should even version-tag the built-in expressions and keep the buggy one when fixing, in this case.

rfcs/0024-nondeterministic-functions.md
xxchan, line 1: Just came up with a problem: IIRC currently UDF is just an expression, which can be used in other places beside Project, e.g., Filter. Should we make them occur only in Project? Besides the determinism problem, there are also optimizations like https://github.com/risingwavelabs/risingwave/pull/8703

xxchan: BTW, https://github.com/risingwavelabs/rfcs/pull/12/files#diff-135315ddd7b7929c8daf439f35043ba3c168ce620ab72c56740a6055d15e8f9fR141 refused to add a new operator, but it seems we eventually need to add some 😅

st1page: Firstly, I need to clarify that the "UDF" and "non-deterministic expression" are independent concepts. `PROC_TIME()` and `RAND()` are built-in expressions but they are non-deterministic. A user-defined function can also be deterministic if a user makes a contract with us.

st1page: And materialized Project is just used for non-deterministic expression on an "updatable stream". We can use a normal project on the append-only stream for non-deterministic expression too.

st1page: I totally agree to refuse to add a new operator for UDF or async expression but not for non-deterministic expression. In another word, if we did add a "UDFProject", now we maybe have to accept `StreamProject`, `StreamAsyncProject`,  `StreamMaterializeProject`, `StreamAsyncMaterializeProject` :sweat_smile: :hot_face:

xxchan: You are right. Non-determinism is not only UDF. 😅

xxchan: Oh, I'm wrong again. I forgot append-only stream. But the points might still apply 😇

xxchan, line 1: > This proposal inspires me that, if we're going to fix a bug for our built-in expression in a version update, it also breaks the determinism, even if the expression itself is mathematically pure.

This reminds me that we have more impure functions than `now()` and `random()`, some of which are not that obvious:

https://github.com/risingwavelabs/risingwave/issues/3552

xxchan: Oh, ignore me. I think they are generally resolved. See comments there.

st1page: I think all the optimization for async execution should be done behind the expression framework 🤔 Well there also are some executor-level optimization. For example, some users might not care about the order of the records on the append-only stream and we might reorder them.

xxchan: > all the optimization for async execution should be done behind the expression framework

Is that enough? If so we can only tweak the performance for expr on one chunk, but not amoung chunks, like `buffered` in  https://github.com/risingwavelabs/risingwave/pull/8703/files

xxchan: Well, actually I don’t mind adding many different executor variants in the future, just like the TopN variants 😅

Let’s conclude the discussion.

rfcs/0024-nondeterministic-functions.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Motivation
+
+In the discussion about the udf, @skyzh raise the issue about the non-deterministic function on updatable stream <https://github.com/risingwavelabs/rfcs/pull/12#issuecomment-1305105614>. This issue also appears on the built-in functions such as `random(),` `proc_time()` or `now()`.
+
+## Design
+
+We will introduce a `MaterializeProject`, it will materialize some **partial** columns of result with the stream key of the input as the primiary key. When an `Insert` operation comes, it will compute the result and materialize some columns. when `Update` or `Delete` comes, it will lookup its state to replace the old value of the operation.
```
xxchan, line 20: Find a problem: not all expressions can be extracted to a `Project` easily, e.g., in `FILTER` clause. 🤡

This is mentioned in https://github.com/risingwavelabs/rfcs/pull/12 as a reason against a dedicated operator for UDF.

PR #26, opened 2022-11-29
RFC: Memory Control

fuyufjh:
Preview: https://github.com/risingwavelabs/rfcs/blob/eric/memory-management/rfcs/0026-memory-control.md

### DIFF
--- a/rfcs/0026-memory-control.md
+++ b/rfcs/0026-memory-control.md
@@ -5,17 +5,27 @@ authors:
 start_date: "2022/11/29"
 ---
 
-# Memory Control
-
-
+# RFC: Memory Management
 
 ## Summary
 
 This RFC proposes a detailed and practical way to control the memory usage for both streaming and batch executors on compute nodes.
 
-## Motivation
+## Background
 
 We had discussed a lot about how to allocate and manage memory before. [RFC: Yet another simple idea for memory management](https://singularity-data.quip.com/CldAAcFmzZSO/Yet-another-simple-idea-for-memory-management) proposed a simple way to control the memory usage of streaming executors by controlling the *watermark epoch* of the LRU caches on a compute node, which was finally adopted in our system, surpassing [RFC: Simpler Dynamic Memory Management Among LRU-Caches](https://singularity-data.quip.com/A1kHAOBUo3Im/RFC-Simpler-Dynamic-Memory-Management-Among-LRU-Caches) and [RFC: Dynamic Memory Budget of Streaming Operators](https://singularity-data.quip.com/J9KYAQc2xIbr/RFC-Dynamic-Memory-Budget-of-Streaming-Operators).
+
+In short, the idea is loop the following steps:
+
+1. **Measure**: Get the total memory usage from `jemalloc` lib. Currently we are using [`tikv_jemalloc_ctl::stats::allocated`](https://docs.rs/tikv-jemalloc-ctl/0.5.0/tikv_jemalloc_ctl/stats/struct.allocated.html), which might be a little bit less than the actual usage. That's why we have introduced "reserved" memory.
+2. **Policy**: The policy will take the current state as well as previous state to decide whether and how much memory should be released. We run policy every barrier interval i.e. 1 second. The difference between previous state and current stats gives the policy a feedback i.e. how much memory has been released since last decision, so it can adjust the policy accordingly. See lastest [`memory_management/policy.rs`](https://github.com/risingwavelabs/risingwave/blob/main/src/compute/src/memory_management/policy.rs) for details.
+3. **Action**: The policy increase a LRU "watermark" i.e. the lowerst epoch of data that should be kept in memory. Streaming executors will evict data with epoch lower than the watermark in proper time.
+
+This method has several drawbacks:
+
+1. It assumes the memory usage of streaming executors dominate the total usage, or in another words, it assumes the memory consumption of storage and batch engine are 0 or some consistand value. Obviously, this is a fragile assumption, and it is never true in practice.
+2. There is no gurantee that the streaming operators can evict data in time aka. in the 1-second feedback loop. The policy assumes all its decision has been applied, but this might not be the truth - some operators cannot react to the new watermark due to many reasons.
+
 
 Recently, [RFC: Coarse grained resource management for batch query engine](https://github.com/risingwavelabs/rfcs/pull/11) tried to control the memory of batch executors as well, but the 1st and 2nd version of the design was controversial because some designs were too complicated or infeasible. Luckily, the RFC has been refined a lot during the discussion, which finally comes to the solution of this RFC. 
 


rfcs/0026-memory-control.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Measure: How to measure memory usage?
+
+Here we propose to use the simple and stupid way to calculate the memory usage - **estimating** the memory consumed by all the in-memory data structures and aggregating in levels. This is opposite to the solutions that rely on memory allocators to get an exact memory usage.
+
+How to estimate memory usage depends on the data structures:
+
+- For indexing structures like `HashMap`, `BTreeMap`, etc., we could estimate the memory size according to the `capacity()` or something else. 
+   - @liurenjie1024: It also seems possible to inject an `Allocator` to count the exact allocated space, as long as it exposed such generic type parameter.
+- For pure data structures like `DataChunk`, `Vec<Datum>`, etc., we could compute the exact size it used by summing up the sizes of internal arrays & variables as well as padding and other overhead.
```
liurenjie1024, line 49: It seems that injecting `Allocator` to `DataChunk` is easier to implement, otherwise we need to override several collections with contains `DataChunk` such as `Vec`, `HeapElement`

rfcs/0026-memory-control.md
```diff
@@ -0,0 +1,113 @@
...
+
+A CN-level memory manager will collect the memory usage in a fixed interval like 100ms. It collects the memory recursively regarding the hierarchy: stream/batch -> fragments -> actors -> executors. To simplify the ownership, executors' memory usage can be stored in `AtomicUsize` and be updated by executors on every `next()`.
```
liurenjie1024, line 51: `AtomicUsize` is too slow, we need to use `hytra`

rfcs/0026-memory-control.md
yuhao-su: It will looks similar to https://github.com/risingwavelabs/risingwave/pull/5932

rfcs/0026-memory-control.md
```diff
@@ -0,0 +1,113 @@
...
+
+Then `free - reserved` is the memory size that could be borrowed by each other, denoted as `overflow`. For example, assuming `limit` of streaming is 500MB and `reserved` is 100MB:
+
+- If `allocated` = 0MB and `free` = 500MB, then `overflow` = 500MB - 100MB = 400MB
+- If `allocated` = 300MB and `free`= 200MB, then `overflow` = 200MB - 100MB = 100MB
+- If `allocated` = 450MB and `free`= 50MB, then `overflow` = 50MB - 100M = -50MB, so overflow is prohibited now
+
+Why `reserved` is necessary? This is because we are using an async way to control memory i.e. with a background coroutine. As a result, **any actions will be slightly later than the actual exhaustion of memory**. The `reserved` memory is designed to mitigate the problem by reserving some space in case that memory cannot be reclaimed immediately.
+
+@fuyufjh prefers the solution without overselling to keep it simple and stupid.
```
liurenjie1024, line 92: I think my diagram is clearer for this case. The algorithm is quite intuitive:
```
if (batch_mem_usage + streaming_mem_usage) >= allowed_mem_usage:
    if batch_mem_usage > (allowed_mem_usage - streaming_mem_reserved):
        reject_new_task();
        kill_batch_task();
    else if batch_mem_usage > batch_mem_reserved:
        reject_new_task();
    else:
        evit_streaming_mem();           
```
A simple illustration for dividing memory areas:
| batch_mem_reserved | dynamic_part | streaming_mem_reserved|

With this simple modification, we can be elastic to more cases and utilize memory better. I really don't understand why this is compliecated.

rfcs/0026-memory-control.md
fuyufjh: > It seems that injecting `Allocator` to `DataChunk` is easier to implement, otherwise we need to override several collections with contains `DataChunk` such as `Vec`, `HeapElement`

Easy but meaningless.

rfcs/0026-memory-control.md
fuyufjh: Yes, I have re-designed this part and I think the complexity is acceptable to me. But I still prefer the "without overselling" solution, because it's super simple! 🤪

rfcs/0026-memory-control.md
liurenjie1024: The complexity is that we need to override container methods, you can think about how to estimate the size of `Vec<DataChunk>`. Anyway it's not that clean.

rfcs/0026-memory-control.md
BowenXiao1999, line 51: Yes I agree. Something like https://github.com/risingwavelabs/rfcs/pull/11#issuecomment-1330349179.

liurenjie1024: I think task level's is enough?

BowenXiao1999: Yes I agree, but the problem is now the executor do not have a way to report memory usage to task level easily (e.g. How frequent executor update the memory usage number?), so it has to come from up to bottom.

liurenjie1024: I think updating an `AtomicUsize` is enough? Passing a task level `AtomicUsize` to all executors.

BowenXiao1999: When will executor update the atomic Usize 🤔

liurenjie1024: That's the reason why I suggest to use `Allocator`: you don't need to care about it, only attach appropriate allocator to necessary part.

BowenXiao1999: But seems like Allocator can not count all usage? 

https://github.com/risingwavelabs/risingwave/blob/ad03b2ef559f5f629313c576f64f626d264863ca/src/stream/src/executor/managed_state/join/join_entry_state.rs#L97-L102

@yuhao-su here is Allocator + estimate size.

liurenjie1024: I guess this is some wrapper around collections such as `BTreeMap`.

jon-chuang: There is already existing work for `TaskLocalAlloc`: https://github.com/risingwavelabs/risingwave/pull/4785

fuyufjh: > When will executor update the atomic Usize 🤔

On every time it processes a DataChunk, I suppose.

fuyufjh:
**Meeting Notes**

- We will do it and start by supporting memory estimation for batch executors, as stated in the "Implementation" section.
- We will work on the "Without Overselling" approach first to see whether it works as expected. Afterwards, we could move to "With Overselling" and tune these parameters

TennyZhuang:
Can we close the RFC?

fuyufjh:
> Can we close the RFC?

I will rewrite the "Measure: How to measure memory usage?" part

PR #27, opened 2022-12-01
RFC: `ManagedRangeCache`'s Last-access-time Datastructure - `RangeLru`

jon-chuang:


0027-managed-range-cache-range-lru.md
```diff
@@ -0,0 +1,58 @@
...
+feature: managed_range_cache_range_lru
+authors:
+  - "jon-chuang"
+start_date: "2022/12/01"
+---
+
+# The Design of `ManagedRangeCache`'s Last-access-time Oracle - `RangeLru`
+
+## Summary
+The design of a mechanism for evicting ranges last accessed before a certain epoch for the `DynamicFilter` operator
```
fuyufjh, line 11: Does the "access" include both writes and reads?

Note that for DynamicFilter the writes are from outer side and the reads are from inner side (i.e. the 1-row side)

0027-managed-range-cache-range-lru.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Details
+
+The dynamic filter requests a given range at every epoch. As this range is the difference between the previous and current pointer, the path of these ranges as a function of the epoch is always continuous.
+
+The current `DynamicFilter` cache, the `RangeCache` differs from other caches in that it caches logical ranges rather than physical elements. That's because the basic unit of storage access for `DynamicFilter` are logical ranges rather than concrete keys.
+
+At any point in time, the `RangeCache` owns a contiguous range.
+
+In-line with our non-OOM and memory-efficiency requirements, we require a way to evict ranges that have not been requested by the operator for a long time. We choose to do so in a similar fashion to `ManagedLru` - based on the last-accessed time of a given subrange (the subranges being a partition of the total range owned by `RangeCache`).
```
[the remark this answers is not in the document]
fuyufjh: Btw, why not maintain the epoch per row?

0027-managed-range-cache-range-lru.md
```diff
@@ -0,0 +1,58 @@
...
+```rust
+/// This updates the last accessed time for a given range. 
+/// `epoch` must be greater than any previously seen epoch.
+fn access(&mut self, range: ScalarRange, epoch: u64)
+
+/// This obtains the ranges subject to deletion in a given epoch
+fn evict_before(&self, epoch: u64) -> Vec<ScalarRange>
+```
+
+1. The main data structure (main index) is a `BTreeMap { epoch -> Vec<ScalarRange> } ` to facilitate the `evict_before` operation.
```
fuyufjh, line 50: Since there should be only one access for each epoch (right?), `Vec` is not needed here

0027-managed-range-cache-range-lru.md
```diff
@@ -0,0 +1,58 @@
...
+2. The secondary index is a modified `BTreeMap`,  storing the left bounds of all the subranges, via the data `BTreeMap { (ScalarImpl, is_exclusive: bool) -> epoch) }`,  allowing one to look up the epochs intersecting with a given range, with an additional field `lower_unbounded_range_epoch: Option<u64>` as lower-unbounded range's existence and epoch cannot be represented. 
+
+Notes:
+1. `evict_before` needs to delete all ranges below the watermark from both the main index and the secondary index
+2. `access` needs to overwrite any range that overlaps with the accessed range in both the main and secondary indexes. In the former, it will modify the endpoints of the ranges that it partially overlaps with, and delete ranges that it fully overlaps with. In the latter, it will delete the left bounds that it fully overlaps with.
```
fuyufjh, line 55: What will happen if the last-access time of subranges are not ordered? For example

```
1.  0 ~ 10  last access epoch = 100
2. 10 ~ 20  last access epoch = 200
3. 20 ~ 30  last access epoch = 150
4. 30 ~ 40  last access epoch = 300
```

We want to `evict_before(160)` and 1 & 3 will be evicted, but 2 & 4 is not continuous subrange.

0027-managed-range-cache-range-lru.md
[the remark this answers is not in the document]
jon-chuang: > Btw, why not maintain the epoch per row?

This would violate the integrity of range cache since it guarantees to cache all rows in a certain range.

0027-managed-range-cache-range-lru.md
fuyufjh: After thinking for a while, I suppose that the "access" should only include reads from inner-side (moving the "row pointer" up or down), because the cache is designed used for reads only. The outer side is basically identical to a `MaterializeExecutor`

fuyufjh:
Please allow me also to propose my idea here. It's too simple to be written into an RFC.

First of all, caching is helpful but not critical to DynamicFilter. I'll show this by discussion in 2 cases:

1. Simple aggregation as the inner side. As we know, the aggregation will only emit the result on every barrier.
2. NowExecutor as the inner side. Similarly, the current time will only be updated on barriers come.

In summary, for major cases of `DynamicFilter`, changes only come on barriers, so it's not as frequent as other operators like HashAgg or HashJoin. Obviously, cache miss per barrier has much less impact on performance.

Second, based on that conclusion, I would prefer either no caching or a very simple caching implementation.

~~(lot of deleted words)~~

After thinking for a while, I felt that even a fixed-size caching is not trivial to implement, so I would like to propose no caching at all.

0027-managed-range-cache-range-lru.md
jon-chuang: Yes, access only indicates reads.

0027-managed-range-cache-range-lru.md
jon-chuang: Every contiguous range of epochs maps onto a contiguous range in the cache. So this scenario will never happen. 

In the impl, we include many sanity checks regarding this continuity property.

0027-managed-range-cache-range-lru.md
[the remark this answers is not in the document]
jon-chuang: > Btw, why not maintain the epoch per row?

Actually, on further thought, perhaps ManagedLru cache can work since it will update all the rows in a range with new timestamp on every read. So it will achieve the same thing but be less efficient since it has to track each row rather than the access of logical range.

0027-managed-range-cache-range-lru.md
fuyufjh: > Every contiguous range of epochs maps onto a contiguous range in the cache

Wow, I suddenly understand what you mean!

yuhao-su: I'm having another case:
```
1. 10~30    epoch = 1
2. 20~30   epoch = 2
3. 20~25    epoch = 3
```
If we evict epoch 1 and 2, then only (20~25) will be kept and (10~20, 25~30) will be evicted. (Not sure if I got this correctly)

jon-chuang: > I'm having another case

Oh, indeed. The range that is evicted is not a continuous range in this case, so I made an error.

So evict before can indeed return two contiguous ranges (the remaining range splitting up the old range).

PR #28, opened 2022-12-01
RFC: DAG Structured Query Plan

chenzl25:


rfcs/0028-dag-structured_query_plan.md
```diff
@@ -0,0 +1,67 @@
...
+
+
+#### Example
+With the `Share` operator we can perform common plan detection at any time of the optimization. I will give some examples:
+
+- CTE. At the binder/planner stage, we can create a `Share` for the table expression and reuse it.
+- Subquery unnesting. We can calculate the domain by creating a `Share` operator of the LHS of the apply operator.
+- Same source operator appears multiple times in a query. Use Share operator at binder/planner or at end of optimizer are both ok.
+- Common sub-plan detection. We can use a field `digest` for each sub-plan and if two sub-plan have the same `digest` it means they are probably the same sub-plan. `digest` is just like a hashcode. We also need to have a `deepEqual` method for each plan to compare with another one.
```
TennyZhuang, line 45: IIUC we can implement `Hash` and `Eq` for them directly.

chenzl25: It depends. Every plan node has its plan id by default, so if we put them into a `HashSet` or other collections, they will be regarded as different item. Once we overwrite this method, the behavior will change.

rfcs/0028-dag-structured_query_plan.md
```diff
@@ -0,0 +1,67 @@
...
+
+![](images/self_join.svg)
```
BugenZhao, line 47: FYI, executing a graph with multiple-edges is currently not a simple deal in the compute node. 🥵
https://www.notion.so/risingwave-labs/Multiple-Edges-in-the-Stream-Graph-f0ab595523d74823a39edaebdea2be16

rfcs/0028-dag-structured_query_plan.md
```diff
@@ -0,0 +1,67 @@
...
+
+The cost of DAG structured plans can be complex if we have a large search space, especially for batch query. Luckily for streaming queries, we don't have a large search space for now and we can always convert the plan to the DAG plan if possible or provide a configuration for users to determine whether to use it or not.
+
+For the batch query, DAG means we need to read the input multiple times which inevitably leads to buffer or materialization. The cost of DAG is determined by how many times it has been read. Normally, the first time has a higher cost than later ones, because of the materialization. It is hard to tell whether to use DAG or not for batch query, I think we can always convert the DAG back to the tree for batch executor.
+
+For the streaming query, it seems not a big deal. We support buffering and dispatch naturally.
+
+### Execution
+
+We only discuss the execution for streaming query here for the reason that batch query needs to do a materialization/buffering which is complicated and hard to tell if DAG plan execution is better or not. `Share` operator has multiple downstream and we can dispatch its data to each downstream just like `ChainNode`.
```
BugenZhao, line 63: FYI, executing a graph with multiple-edges is currently not a simple deal in the compute node. 🥵
https://www.notion.so/risingwave-labs/Multiple-Edges-in-the-Stream-Graph-f0ab595523d74823a39edaebdea2be16

chenzl25: Yes, I remember this issue. Based on this RFC, if a node has multiple downstream, it must be the `Share` node.  The fragmenter can split the `Share` node just like it splits the exchange node. In this way, we can separate the upstream and the downstream into different fragments. So the last thing we need to deal with is ensuring all the downstreams of the `Share` Node are in different fragments. If there multiple edges are connected to the same downstream fragment, use shuffle to separate them.

rfcs/0028-dag-structured_query_plan.md
```diff
@@ -0,0 +1,67 @@
...
+
+Last but not least, at the end of the optimization, we might find that the `Share` operator has only one parent. In this case, we can remove the `Share` operator.
+
+#### Cost
+
+The cost of DAG structured plans can be complex if we have a large search space, especially for batch query. Luckily for streaming queries, we don't have a large search space for now and we can always convert the plan to the DAG plan if possible or provide a configuration for users to determine whether to use it or not.
+
+For the batch query, DAG means we need to read the input multiple times which inevitably leads to buffer or materialization. The cost of DAG is determined by how many times it has been read. Normally, the first time has a higher cost than later ones, because of the materialization. It is hard to tell whether to use DAG or not for batch query, I think we can always convert the DAG back to the tree for batch executor.
+
+For the streaming query, it seems not a big deal. We support buffering and dispatch naturally.
```
fuyufjh, line 59: Interesting observation. Indeed, for streaming, execution of DAG plan doesn't introduce any additional cost. No extra buffering needed.

xxchan (commented):
Completed at https://github.com/risingwavelabs/risingwave/issues/6955

xxchan (approved):

PR #29, opened 2022-12-05
RFC: CDC Source Node (MySQL)

StrikeW:
Migrated from Notion.


tabVersion:
will we have another discussion or merge directly?

StrikeW:
> will we have another discussion or merge directly?

It is just a migration of the previous RFC. The conclusion is below.
- To make the source node pure stateless, its lifecycle should align with compute node (source executors). When the CN crashed the source node will experience a grpc connection error, then corresponding CDC engines should exit.
- Based on above conclusion, we can use only one rpc interface to create and consume CDC events
- The source (connector) node will deploy as a sidecar with compute node
- To verify the database connector properties we may need to deploy source node as sidecar with Meta node and verify in the process of create source.

xxchan (approved):
Let's merge it.

>  The conclusion is below.

Does this need to me edited into the RFC?

PR #31, opened 2022-12-18
s3 source with csv parser

waruto210:
Support load data from Amazon S3 and other Simple Storage Services with CSV format.

### DIFF
--- a/rfcs/0000-s3-source.md
+++ b/rfcs/0000-s3-source.md
@@ -1,0 +1,109 @@
+---
+feature: S3 source with csv parser
+authors:
+  - "Mingchao Wu"
+start_date: "2022/11/23"
+---
+
+# My Excited Feature
+
+## Summary
+
+Support load data from Amazon S3 and other Simple Storage Services with CSV format.
+
+
+## Design
+
+To implement the S3 data source, an abstraction of the file system data source is proposed.
+
+### Overall Process
+
+`Split Enumerator`: Continuously list all files/objects in the directory/bucket
+
+`FsSourceExecutor`:
+
+1. get all splits from barrier mutation
+2. scan info of all finished splits from the state store, then filter out finished splits
+3. start reading the filtered splits
+4. when taking a snapshot, a special prefix is used to identify the finished split.
+
+### Implementation
+
+- Split
+
+  ```rust
+  **pub struct FsSplit {
+      pub name: String,
+      pub offset: usize,
+      pub size: usize,
+  }**
+  ```
+
+- Split reader
+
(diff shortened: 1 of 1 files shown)


rfcs/0031-s3-source.md
```diff
@@ -0,0 +1,110 @@
...
+      pub split_size: usize,
+      pub split_id: SplitId,
+  }
+  ```
+
+- Parser
+
+  A file is different from a message queue, where a file is read as a series of byte chunks, instead of separate messages.
+
+  The parser should try to parse one record from the payload, if the payload is not enough to parse one record, the parser will buffer the payload in its internal buffer and wait for more bytes to be passed. If any error occurs, the parser should clean its internal buffer.
```
fuyufjh, line 69: +1 for the "statefulness". In the future, I suppose the parser should be stateful, not only for buffering some input bytes, but also for storing some necessary metadata read from file's header (like ORC or Parquet files). Hopefully, this design can live up to those requirements.

rfcs/0031-s3-source.md
```diff
@@ -0,0 +1,110 @@
...
+      /// - writer: Write exactly one record during a `parse` call.
+      ///
+      /// # Returns
+      ///
+      /// A [`WriteGuard`] to ensure that at least one record was appended or error occurred
+      fn parse<'a, 'b, 'c>(
+          &'a mut self,
+          payload: &'a mut &'b [u8],
+          writer: SourceStreamChunkRowWriter<'c>,
+      ) -> Self::ParseResult<'a>
```
fuyufjh, line 91: By @hzxa21: May consider implementing it like a _streaming operator_ to input streaming bytes and output streaming events. By the way, our streaming code adopts `futures_async_stream` to do this, which basically allows you to `yield` a DataChunk any time and the compiler & macro will handle other things like how to keep the states.

https://github.com/risingwavelabs/risingwave/blob/0128ff1b312ae22f51e979a73bd5cb9f61831c39/src/stream/src/executor/project_set.rs#L196

rfcs/0031-s3-source.md
fuyufjh: In addition, in this future, we should use **composing** parser inside reader to make it more flexible. In this cases, NexMark connectors can directly output DataChunk without the redundant serialization.

rfcs/0031-s3-source.md
```diff
@@ -0,0 +1,110 @@
+---
+feature: S3 source with csv parser
+authors:
+  - "Mingchao Wu"
+start_date: "2022/11/23"
+---
+
+# My Excited Feature
```
fuyufjh, line 8:
```suggestion
# S3 source with csv parser
```

PR #32, opened 2022-12-26
RFC: Band Join

chenzl25:


yuhao-su:
The implementation proposed in this RFC may enlighten a way to [Temporal join](https://nightlies.apache.org/flink/flink-docs-release-1.16/docs/dev/table/sql/queries/joins/#temporal-joins). 

Also, it is possible to implement with shared arrangement. (Maybe can avoid broadcast)

chenzl25:
> The implementation proposed in this RFC may enlighten a way to [Temporal join](https://nightlies.apache.org/flink/flink-docs-release-1.16/docs/dev/table/sql/queries/joins/#temporal-joins).
> 
> Also, it is possible to implement with shared arrangement. (Maybe can avoid broadcast)

I think actually it is more related to [Interval Joins](https://nightlies.apache.org/flink/flink-docs-release-1.16/docs/dev/table/sql/queries/joins/#interval-joins).

rfcs/0032-band-join.md
```diff
@@ -0,0 +1,81 @@
...
+
+```sql
+-- This is a band-join.
+select * from A join B on A.p between B.d - 10 and B.d + 20.
+
+-- This is not a band-join, but a range join.
+select * from A join B on A.p between B.start and B.end.
+```
+
+The band join has a nice property. The range condition of band join only involves 2 columns: one from the LHS the other from RHS, so we can always reverse the condition. For example, `A.p between B.d - 10 and B.d + 20` can be converted into `B.d between A.p - 20 and A.p + 10`. This is a crucial property for streaming queries, as we need to treat both sides of the join logically equivalent (both sides need to be built and probed).
```
CAJan93, line 20: > This is a crucial property for streaming queries, as we need to treat both sides of the join logically equivalent (both sides need to be built and probed). 


Why do we need to probe both sides? Why is this different in streaming vs. batch?

rfcs/0032-band-join.md
```diff
@@ -0,0 +1,81 @@
...
+### Streaming
+
+![](./images/0032-band-join/band-join.png)
+
+For simplicity, we consider the following query first.
+```sql
+select * from A join B on A.p between B.d - 10 and B.d + 20.
+```
+
+- We can use singleton for both input sides or broadcast one side to the other side. We will construct an internal table for A side with order key = `A.p, A.rid` and for B side with order key = `B.d, B.rid`.
```
CAJan93, line 38: What does the `rid` mean in `A.rid` and `B.rid`?

CAJan93: Row Identifier. Okay. Never mind :D

rfcs/0032-band-join.md
```diff
@@ -0,0 +1,81 @@
...
+- We can reverse the condition `A.p between B.d - 10 and B.d + 20` into `B.d between A.p - 20 and Ap.p + 10`.
+- Suppose we already have 3 rows for A side (10, 10001), (20, 10002), (30, 10003) and 3 rows for B side (5, 20001), (15, 20002), (45, 20003).
+- When A side row changed comes with +(15, 10004), we need to search B side with condition `B.d between 15 - 20 = -5 and 15 + 10 = 25`, so B side row (5 , 20001) and (15, 20002) matched and return.
+- When A side row change comes with -(20, 10002), we need to search B side with condition `B.d between 20 - 20 = 0 and 20 + 10 = 30`, so B side row (5 , 20001) and (15, 20002) matched and return.
```
CAJan93, line 42: I do not get the change in `A`. Does `+(15, 10004)` mean that we add 15 to the value of the RID 10004. If so this is not really an update as an append, right? Since we did not see these RIDs before? Please tell me if I misinterpret this.

rfcs/0032-band-join.md
```diff
@@ -0,0 +1,81 @@
...
+## Unresolved questions
+
+* Parallelize the streaming `BandJoin` if there is no equal condition. Some papers have discussed how to optimize the distributed `BandJoin` by partitioning the input datasets, but it might need a special shuffle algorithm.
+* For simplicity, we can only support `InnerJoin` for now.
+
+## Future possibilities
+
+If you are familiar with Flink, we can find that they have interval join which is just a special case of the `BandJoin`. Interval join requires an equal condition and the range condition looks like `b.timestamp ∈ [a.timestamp + lowerBound; a.timestamp + upperBound]`. Interval join also requires the input stream to be append-only and can cooperate with the watermark to prune the old states. We can use `BandJoin` to implement interval join in the future.
+
+## Reference
```
CAJan93, line 73: Thank you for including the references. That helped :D

rfcs/0032-band-join.md
chenzl25: For streaming, updates from either left side or right side need to be reflected on the join output. Think about `StreamHashJoin`, we need to build a hash table for both sides and if an update comes from one side, we need to use it to probe the other side to find the matched rows. This is also applicable to the stream-band-join. As you can see, we need to  build some index structures for both sides in streaming, while in batch we just need to build an index structure for one side (the build side), because the input size is bounded in batch.

rfcs/0032-band-join.md
chenzl25: `+(15, 10004)` means it is an insertion(+) of a row `(15, 10004)` with A.p = 15 and A.rid = 10004.

rfcs/0032-band-join.md
```diff
@@ -0,0 +1,81 @@
...
+- If we don't have equal conditions `A.a = B.b`, we can only use singleton for both input sides or broadcast one side to the other side. The other logic is basically equivalent with the above example without `A.a` and `B.b` as their prefix keys.
```
liurenjie1024, line 55: We should be careful when doing broadcast since it duplicates the broadcasted state table. It would be better to do it with user hint.

rfcs/0032-band-join.md
```diff
@@ -0,0 +1,81 @@
...
+- If we have an equal condition `A.a = B.b` and its selectivity is high, we can use this condition to distribute the data to acquire parallelism. For `A.p between B.d - 10 and B.d + 20`, we can construct an internal table with order key = `A.a, A.p, A.rid` for A side and order key `B.b, B.d, B.rid` for B side. For `A.q between B.e - 10 and B.e + 20`, we can construct an internal table with order key = `A.a, A.q, A.rid` for A side and order key `B.b, B.e, B.rid` for B side. When a row came from B with (B.b, B.d, B.e) = (100, 200, 300). We can lookup A's internal table row ids with range queries: A between (A.a = 100, A.p = 200 - 10 = 190) and (A.a = 100, A.p = 200 + 20 = 220). Merge the other A's internal table row ids with range queries: A between (A.a = 100, A.q = 300 - 5 = 295) and (A.a = 100, A.p = 200 + 15 = 315). Finally we can intersect the row ids to get the corresponding A matched rows. When a row comes from A, we first need to reverse the range condition as we mentioned before and then do the same logic as for row came from B. Row deleted is basically equivalent to the insertion, but with opposed operators. Update can be handled as delete followed by insert.
```
fuyufjh, line 54: In streaming I guess most band join conditions are on timestamp columns, so it seems fine to only support one band condition.

rfcs/0032-band-join.md
```diff
@@ -0,0 +1,81 @@
...
+
+
+## Unresolved questions
+
+* Parallelize the streaming `BandJoin` if there is no equal condition. Some papers have discussed how to optimize the distributed `BandJoin` by partitioning the input datasets, but it might need a special shuffle algorithm.
+* For simplicity, we can only support `InnerJoin` for now.
+
+## Future possibilities
+
+If you are familiar with Flink, we can find that they have interval join which is just a special case of the `BandJoin`. Interval join requires an equal condition and the range condition looks like `b.timestamp ∈ [a.timestamp + lowerBound; a.timestamp + upperBound]`. Interval join also requires the input stream to be append-only and can cooperate with the watermark to prune the old states. We can use `BandJoin` to implement interval join in the future.
```
fuyufjh, line 71: Consider using the time-bound condition to do state cleaning, there is a unresolve problem: should the time column be placed before join key or after join key?

1. time column before join key: state cleanning is easy (one range delete), but random acesss by join key become impossible.
2. join key before time column: Join key access is fast as usual, but state cleanning's cost is O(N) where N = number of join keys.
3. Alternatively, use two seperated state tables, but this would increase the IO cost multiple times.

Neither sounds a good solution. We need to take a look on Flink's implementation.

rfcs/0032-band-join.md
chenzl25: Yes, if there are more than one band condition, we can always choose one of them to construct the internal state and other band conditions can be just treated as other conditions.

rfcs/0032-band-join.md
chenzl25: I had taken a look at Flink's `TimeIntervalJoin` and found that they use alternative two. First they utilize the join equal keys to partition the join state. Second, they maintain a `leftCache` and `rightCache` with type `MapState<Long, List<Tuple2<RowData, Boolean>>>` for each join equal key. The `Long` type is actually the timestamp. BTW, they suffer from a lack of ordered key MapState. The state cleaning mechanism of Flink `TimeIntervalJoin` works like that, it registers a cleanup timer to its time service. The cleanup timer triggered timestamp is calculated based on the row timestamp, input watermark and time interval provided by the user. As soon as the cleanup timer triggers, it iterates the LeftCache and RightCache, deletes expired rows, and registers another cleanup timer based on the earliest valid row timestamp left. As we can see, 
Flink uses a bunch of point deletes triggered by the cleanup timer, rather than a range delete.

chenzl25: https://github.com/apache/flink/blob/master/flink-table/flink-table-runtime/src/main/java/org/apache/flink/table/runtime/operators/join/interval/TimeIntervalJoin.java

soundOfDestiny: > Consider using the time-bound condition to do state cleaning, there is a unresolve problem: should the time column be placed before join key or after join key?
> 
> 1. time column before join key: state cleanning is easy (one range delete), but random acesss by join key become impossible.
> 2. join key before time column: Join key access is fast as usual, but state cleanning's cost is O(N) where N = number of join keys.
> 3. Alternatively, use two seperated state tables, but this would increase the IO cost multiple times.
> 
> Neither sounds a good solution. We need to take a look on Flink's implementation.

I think the first solution is ridiculous. We do not need a fast state cleanning.

fuyufjh (approved):
Shall we merge this?

chenzl25:
> Shall we merge this?

Yes, because we have support `Interval Join` which could clean the state at least.

PR #34, opened 2023-01-12
RFC: System Parameters on Meta

fuyufjh:
Preview: https://github.com/risingwavelabs/rfcs/blob/eric/system-parameter/rfcs/0034-system-parameters.md

Related https://github.com/risingwavelabs/risingwave/issues/7203

### DIFF
--- a/rfcs/0034-system-parameters.md
+++ b/rfcs/0034-system-parameters.md
@@ -22,7 +22,7 @@ These can not meet the requirements for several reasons.
 
 This RFC proposes to introduce centralized configurations on Meta Service and data will be persisted in etcd.
 
-For the sake of clarity, in this doc, we will use the word "parameters" for system and session parameters, and "config" for command-line arguments or config files.
+For the sake of clarity, in this doc, we will use the term **"parameters"** for system and session parameters, and **"config"** for command-line arguments or config files.
 
 **Note:** Previously I think this might not be an urgent issue until I realized that our Cloud platform couples tightly with more and more configurations. The later we commence, the heavier burden it will become. 
 
@@ -53,7 +53,7 @@ ALTER SYSTEM / DATABASE SET parameter TO value / 'value' / DEFAULT;
 SHOW PARAMETERS;
 ```
 
-> NOTE: PG uses `SHOW ALL` because it doesn't has other `SHOW` commands like us. Which one is better?
+> **Note**: PG uses `SHOW ALL` because it doesn't has other `SHOW` commands like us. Which one is better?
 
 
 References:
@@ -81,13 +81,13 @@ References:
 
 - [PostgreSQL: Documentation: 15: 20.1. Setting Parameters](https://www.postgresql.org/docs/current/config-setting.html)
 
-Previously, we had already introduced [session parameters in RisingWave](https://github.com/risingwavelabs/risingwave/blob/53f7e0db772ac7e51773791bb8301624ed763ae8/src/common/src/session_config/mod.rs#L265). Some of them can also be system parameters like `batch_enable_sort_agg` or `query_mode`. Postgres enforces that session parameters must also be system paramters, we can follow this rule as well.
+Previously, we had already introduced [session parameters in RisingWave](https://github.com/risingwavelabs/risingwave/blob/53f7e0db772ac7e51773791bb8301624ed763ae8/src/common/src/session_config/mod.rs#L265). Some of them should be system parameters like `batch_enable_sort_agg` or `query_mode`. Postgres **enforces** that session parameters must also be system paramters, we can follow this rule as well.
 
 ### Process Configs
 
 Let's rethink the current design of CLI options and config files. ([#5676](https://github.com/risingwavelabs/risingwave/issues/5676)) Given the existence of system parameters, some existing parameters may need to be moved to system parameters, like `data_directory` as mentioned before.
 
-We propose to **treat** CLI options as a way to override config files, as a result, the items in CLI options are a subset of the config file. The reasons for this include
+We propose to consider CLI options as a way to override config files (i.e. in term of priority, CLI options -> config file -> defaults); as a result, the items in CLI options are a subset of the config file. The reasons for this include
 
 - All configs are still too many to be written as CLI options. (See [example.toml](https://github.com/risingwavelabs/risingwave/blob/main/src/config/example.toml)) 
 - While CLI configs are handy to set some frequently-used options like endpoint and memory.
@@ -130,3 +130,6 @@ block_size_kb = 1024
 - Global configs should be flatten unlike TOML. Consider add a prefix e.g. `stream_foo`, `batch_foo` if it seems ambiguous.
 - Dangerous configs should be named with keyword `developer` or `unsafe` as a warning.
 
+## Discussions
+
+1. Do we need database parameters, global (instance-wide) parameters, or both?
--- a/rfcs/0034-system-parameters.md
+++ b/rfcs/0034-system-parameters.md
@@ -5,27 +5,26 @@ authors:
 start_date: "2023/01/11"
 ---
 
-# RFC: System Parameters on Meta
+# RFC: System Parameters
 
 ## Motivation
 
-Currently, we have 2 kinds of configurations:
+Currently, we have 3 kinds of configurations:
 
 - Command-line arguments
 - Configuration file `risingwave.toml`
+- Session parameters
 
-[Distinguish system config variables and cli options #5676](https://github.com/risingwavelabs/risingwave/issues/5676) clarifies the difference between CLI args and config file:
+These can not meet the requirements for several reasons.
 
-> Intuitively, *Options* should contain basic user-facing options, e.g., addr & host and maybe some feature gates, while *Config* contains deep-dark internal system variables to fine-tune the system performance.
+1. **Consistency**. `risingwave.toml` is a local file, which makes it possible that multiple nodes could see different parameters and this is absolutely unacceptable. For example, if some nodes use `data_directory = 'hummock_001'` and others use `data_directory = 'hummock_002'`, the stored data will be corrupted.
+2. **Mutability**. Users are unable to `SET` a config after the process started. The root cause is still the local file - changes of configs must be persisted in some cluster-level storage.
 
-However, this is not a complete solution because
+This RFC proposes to introduce centralized configurations on Meta Service and data will be persisted in etcd.
 
-1. **`risingwave.toml` is a local file**, which makes it possible that multiple nodes could see different parameters and this is absolutely unacceptable. For example, if some nodes use `data_directory = 'hummock_001'` and others use `data_directory = 'hummock_002'`, the stored data will be corrupted.
-2. **Unable to `SET` a config after the process started**. Even if changes are written to disk, new nodes still miss these changes.
+For the sake of clarity, in this doc, we will use the word "parameters" for system and session parameters, and "config" for command-line arguments or config files.
 
-This proposal will continue to improve the configuration mechanism but also repudiate the design of #5676.
-
-Previously I think this might not be an urgent issue until I realized that our Cloud platform couples tightly with more and more configurations. The later we commence, the heavier burden it will become. 
+**Note:** Previously I think this might not be an urgent issue until I realized that our Cloud platform couples tightly with more and more configurations. The later we commence, the heavier burden it will become. 
 
 ## Design
 
@@ -33,14 +32,14 @@ Previously I think this might not be an urgent issue until I realized that our C
 graph LR
     all[Does it apply to the RisingWave instance?]
     all ==No==> process["Process Configs (Local)"]
-    all ==Yes==> system["System Configs (on Meta)"]
+    all ==Yes==> system["System Parameters (on Meta)"]
     process --> cli_args(Command-line arguments)
     process --> conf_file("Config file (.toml)")
     system --> sql_cmd(SQL interface: SET / SHOW)
     system --> cloud(Web Console)
 ```
 
-### System Configs
+### System Parameters
 
 What we need is a centralized source-of-truth of cluster configurations. Obviously, the best choice for RisingWave is the Meta Service and the underlying etcd. 
 
@@ -57,7 +56,7 @@ SHOW PARAMETERS;
 > NOTE: PG uses `SHOW ALL` because it doesn't has other `SHOW` commands like us. Which one is better?
 
 
-See also: 
+References:
 
 - [PostgreSQL: Documentation: 15: ALTER SYSTEM](https://www.postgresql.org/docs/current/sql-altersystem.html)
 - [PostgreSQL: Documentation: 15: 20.1. Setting Parameters](https://www.postgresql.org/docs/current/config-setting.html)
@@ -70,15 +69,34 @@ Similar to other system metadata e.g. catalog,
 
 Therefore, the Notification Service seems to fit the scene well.
 
-### Process-level Configs
+### Session Parameters
 
-Given the existance of System Config, how to design the machine/process-level CLI options and config file? We propose to **unify** them because local parameters are still too many to be written as CLI options. (See [example.toml](https://github.com/risingwavelabs/risingwave/blob/main/src/config/example.toml)) 
+PostgreSQL allows users to set parameters for the current session as well.
+
+```sql
+SET parameter TO value / 'value' / DEFAULT;
+```
+
+References:
+
+- [PostgreSQL: Documentation: 15: 20.1. Setting Parameters](https://www.postgresql.org/docs/current/config-setting.html)
+
+Previously, we had already introduced [session parameters in RisingWave](https://github.com/risingwavelabs/risingwave/blob/53f7e0db772ac7e51773791bb8301624ed763ae8/src/common/src/session_config/mod.rs#L265). Some of them can also be system parameters like `batch_enable_sort_agg` or `query_mode`. Postgres enforces that session parameters must also be system paramters, we can follow this rule as well.
+
+### Process Configs
+
+Let's rethink the current design of CLI options and config files. ([#5676](https://github.com/risingwavelabs/risingwave/issues/5676)) Given the existence of system parameters, some existing parameters may need to be moved to system parameters, like `data_directory` as mentioned before.
+
+We propose to **treat** CLI options as a way to override config files, as a result, the items in CLI options are a subset of the config file. The reasons for this include
+
+- All configs are still too many to be written as CLI options. (See [example.toml](https://github.com/risingwavelabs/risingwave/blob/main/src/config/example.toml)) 
+- While CLI configs are handy to set some frequently-used options like endpoint and memory.
 
 This pattern is widely adopted in many databases and even other applications. For example, TiDB considers configuration file as a superset of CLI options:
 
-> The TiDB configuration file supports more options than command-line parameters. You can download the default configuration file `[config.toml.example](https://github.com/pingcap/tidb/blob/master/config/config.toml.example)` and rename it to `config.toml`. This document describes only the options that are not involved in [command line options](https://docs.pingcap.com/tidb/stable/command-line-flags-for-tidb-configuration).
+> The TiDB configuration file supports more options than command-line parameters. You can download the default configuration file [config.toml.example](https://github.com/pingcap/tidb/blob/master/config/config.toml.example) and rename it to `config.toml`. This document describes only the options that are not involved in [command line options](https://docs.pingcap.com/tidb/stable/command-line-flags-for-tidb-configuration).
 >
-> 🔗 [TiDB Configuration File | PingCAP Docs](https://docs.pingcap.com/tidb/stable/tidb-configuration-file)
+> Ref. [TiDB Configuration File | PingCAP Docs](https://docs.pingcap.com/tidb/stable/tidb-configuration-file)
 
 In RisingWave it will be like
 
@@ -97,7 +115,6 @@ block_size_kb = 1024
 # Override total_memory_bytes as 2GB
 ./risingwave compute-node --total-memory-bytes=2147483648
 ```
-
 
 ### Mutability
 
--- a/rfcs/0034-system-parameters.md
+++ b/rfcs/0034-system-parameters.md
@@ -1,11 +1,11 @@
 ---
-feature: system-settings
+feature: system-parameters
 authors:
   - "Eric Fu"
 start_date: "2023/01/11"
 ---
 
-# RFC: System Settings on Meta
+# RFC: System Parameters on Meta
 
 ## Motivation
 
@@ -20,7 +20,7 @@ Currently, we have 2 kinds of configurations:
 
 However, this is not a complete solution because
 
-1. `**risingwave.toml` is a local file**, which makes it possible that multiple nodes see different parameters and this absolutely unacceptable. For example, some nodes use `data_directory = 'hummock_001'` and others use `data_directory = 'hummock_002'`, the cluster must be crashed.
+1. **`risingwave.toml` is a local file**, which makes it possible that multiple nodes see different parameters and this absolutely unacceptable. For example, some nodes use `data_directory = 'hummock_001'` and others use `data_directory = 'hummock_002'`, the cluster must be crashed.
 2. **Unable to `SET` a config after the process started**. Even if changes are written to disk, new nodes still miss these changes.
 
 This proposal will continue to improve the configuration mechanism but also repudiate the design of #5676.
--- a/rfcs/0034-system-parameters.md
+++ b/rfcs/0034-system-parameters.md
@@ -20,7 +20,7 @@ Currently, we have 2 kinds of configurations:
 
 However, this is not a complete solution because
 
-1. **`risingwave.toml` is a local file**, which makes it possible that multiple nodes see different parameters and this absolutely unacceptable. For example, some nodes use `data_directory = 'hummock_001'` and others use `data_directory = 'hummock_002'`, the cluster must be crashed.
+1. **`risingwave.toml` is a local file**, which makes it possible that multiple nodes could see different parameters and this is absolutely unacceptable. For example, if some nodes use `data_directory = 'hummock_001'` and others use `data_directory = 'hummock_002'`, the stored data will be corrupted.
 2. **Unable to `SET` a config after the process started**. Even if changes are written to disk, new nodes still miss these changes.
 
 This proposal will continue to improve the configuration mechanism but also repudiate the design of #5676.
@@ -83,7 +83,7 @@ This pattern is widely adopted in many databases and even other applications. Fo
 In RisingWave it will be like
 
 ```toml
-// risingwave.toml
+# risingwave.toml
 [server]
 total_memory_bytes = 1073741824
 parallelism = 4
--- a/rfcs/0034-system-parameters.md
+++ b/rfcs/0034-system-parameters.md
@@ -30,12 +30,12 @@ For the sake of clarity, in this doc, we will use the term **"parameters"** for 
 
 ```mermaid
 graph LR
-    all[Does it apply to the RisingWave instance?]
+    all[Does it apply to multiple components in the cluster or needs to be mutable at runtime?]
     all ==No==> process["Process Configs (Local)"]
     all ==Yes==> system["System Parameters (on Meta)"]
     process --> cli_args(Command-line arguments)
     process --> conf_file("Config file (.toml)")
-    system --> sql_cmd(SQL interface: SET / SHOW)
+    system --> sql_cmd(SQL interface: ALTER / SHOW)
     system --> cloud(Web Console)
 ```
 
@@ -69,25 +69,11 @@ Similar to other system metadata e.g. catalog,
 
 Therefore, the Notification Service seems to fit the scene well.
 
-### Session Parameters
-
-PostgreSQL allows users to set parameters for the current session as well.
-
-```sql
-SET parameter TO value / 'value' / DEFAULT;
-```
-
-References:
-
-- [PostgreSQL: Documentation: 15: 20.1. Setting Parameters](https://www.postgresql.org/docs/current/config-setting.html)
-
-Previously, we had already introduced [session parameters in RisingWave](https://github.com/risingwavelabs/risingwave/blob/53f7e0db772ac7e51773791bb8301624ed763ae8/src/common/src/session_config/mod.rs#L265). Some of them should be system parameters like `batch_enable_sort_agg` or `query_mode`. Postgres **enforces** that session parameters must also be system paramters, we can follow this rule as well.
-
 ### Process Configs
 
 Let's rethink the current design of CLI options and config files. ([#5676](https://github.com/risingwavelabs/risingwave/issues/5676)) Given the existence of system parameters, some existing parameters may need to be moved to system parameters, like `data_directory` as mentioned before.
 
-We propose to consider CLI options as a way to override config files (i.e. in term of priority, CLI options -> config file -> defaults); as a result, the items in CLI options are a subset of the config file. The reasons for this include
+We propose to consider CLI options as a way to override config files (i.e. in term of priority, CLI options -> config file -> defaults). The reasons for this include
 
 - All configs are still too many to be written as CLI options. (See [example.toml](https://github.com/risingwavelabs/risingwave/blob/main/src/config/example.toml)) 
 - While CLI configs are handy to set some frequently-used options like endpoint and memory.
--- a/rfcs/0034-system-parameters.md
+++ b/rfcs/0034-system-parameters.md
@@ -17,7 +17,7 @@ Currently, we have 3 kinds of configurations:
 
 These can not meet the requirements for several reasons.
 
-1. **Consistency**. `risingwave.toml` is a local file, which makes it possible that multiple nodes could see different parameters and this is absolutely unacceptable. For example, if some nodes use `data_directory = 'hummock_001'` and others use `data_directory = 'hummock_002'`, the stored data will be corrupted.
+1. **Consistency**. `risingwave.toml` is a local file, which makes it possible that multiple nodes could see different parameters and this is absolutely unacceptable. For example, to configure the remote directory for storing data, users may use the `data_directory` parameter. If some nodes use `data_directory = 'hummock_001'` and others use `data_directory = 'hummock_002'`, the stored data will be corrupted.
 2. **Mutability**. Users are unable to `SET` a config after the process started. The root cause is still the local file - changes of configs must be persisted in some cluster-level storage.
 
 This RFC proposes to introduce centralized configurations on Meta Service and data will be persisted in etcd.
--- a/rfcs/0034-system-settings.md
+++ b/rfcs/0034-system-settings.md
@@ -1,0 +1,115 @@
+---
+feature: system-settings
+authors:
+  - "Eric Fu"
+start_date: "2023/01/11"
+---
+
+# RFC: System Settings on Meta
+
+## Motivation
+
+Currently, we have 2 kinds of configurations:
+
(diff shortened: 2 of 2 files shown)


Gun9niR:
What if some system configs are required on cluster start, e.g. (hummock URL)?

fuyufjh:
> What if some system configs are required on cluster start, e.g. (hummock URL)?

Initialize it until the parameters are fetched from Meta, or `set` it afterwards 😇 I have no better ideas than these.

Gun9niR:
> Initialize it until the parameters are fetched from Meta

It works for worker types other than Meta, but what about Meta itself? How should these configs be initialized, maybe in command line options?

fuyufjh:
> > Initialize it until the parameters are fetched from Meta
> 
> It works for worker types other than Meta, but what about Meta itself? How should these configs be initialized, maybe in command line options?

How about setting it before any compute node join the cluster?

CAJan93:
Regarding the decision where to put an argument: If e.g. the argument relates to a RW instance, how do we decide if we put it into the SQL interface or the web console? Or will we always offer both options? 

Same for process configs: When do we use a CLI arg and when the .toml file?

xxchan:
> how do we decide if we put it into the SQL interface or the web console?
>
> When do we use a CLI arg and when the .toml file?

IIUC it's layered config:
- web console is subset of SQL variables.
- CLI args is subset of config file.

When some configs are frequently used, we can provide an alternative interface for ease of use.

rfcs/0034-system-parameters.md
```diff
@@ -0,0 +1,115 @@
...
+- Configuration file `risingwave.toml`
+
+[Distinguish system config variables and cli options #5676](https://github.com/risingwavelabs/risingwave/issues/5676) clarifies the difference between CLI args and config file:
+
+> Intuitively, *Options* should contain basic user-facing options, e.g., addr & host and maybe some feature gates, while *Config* contains deep-dark internal system variables to fine-tune the system performance.
+
+However, this is not a complete solution because
+
+1. **`risingwave.toml` is a local file**, which makes it possible that multiple nodes could see different parameters and this is absolutely unacceptable. For example, if some nodes use `data_directory = 'hummock_001'` and others use `data_directory = 'hummock_002'`, the stored data will be corrupted.
+2. **Unable to `SET` a config after the process started**. Even if changes are written to disk, new nodes still miss these changes.
```
xxchan, line 24: i.e.,
1. consistency
2. mutability
🤔

rfcs/0034-system-parameters.md
```diff
@@ -0,0 +1,115 @@
...
+
+This proposal will continue to improve the configuration mechanism but also repudiate the design of #5676.
+
+Previously I think this might not be an urgent issue until I realized that our Cloud platform couples tightly with more and more configurations. The later we commence, the heavier burden it will become. 
+
+## Design
+
```
xxchan, line 31: To summarize, I think it's two designs:
- Distinguish system (cluster) configs from process configs. This is to resolve the main concerns in the motivation section.
- Introduce layered config to replace separated configs. (e.g., CLI args become subset of config file. Instead of two distinct set of configs.) Although LGTM, it doesn't seem to belong to any problem described in motivation 🤪

rfcs/0034-system-parameters.md
```diff
@@ -0,0 +1,115 @@
...
+    all[Does it apply to the RisingWave instance?]
+    all ==No==> process["Process Configs (Local)"]
+    all ==Yes==> system["System Configs (on Meta)"]
+    process --> cli_args(Command-line arguments)
+    process --> conf_file("Config file (.toml)")
+    system --> sql_cmd(SQL interface: SET / SHOW)
+    system --> cloud(Web Console)
+```
+
+### System Configs
```
xxchan, line 43: I think we need to clarify: Are system configs distinct from config files? From my understanding of the motivation it's yes. A related question is: If so, the default values for them (i.e., the values at startup) cannot be changed, right? 🤔

But since you referred PG, in PG they are the same set of configs:

> These variables can be set using the `SET` statement, by editing the `postgresql.conf` configuration file, through the `PGOPTIONS` environmental variable (when using libpq or a libpq-based application), or through command-line flags when starting the `postgres` server.

rfcs/0034-system-parameters.md
```diff
@@ -0,0 +1,115 @@
...
+
+What we need is a centralized source-of-truth of cluster configurations. Obviously, the best choice for RisingWave is the Meta Service and the underlying etcd. 
+
+Here we could borrow the syntax from Postgres:
+
+```sql
+-- Set or reset a parameter
+ALTER SYSTEM / DATABASE SET parameter TO value / 'value' / DEFAULT;
```
xxchan, line 51: Another thing not discussed is session variables. `SET [ SESSION | LOCAL ] configuration_parameter` affects local setting, while `ALTER SYSTEM SET` affects global setting. From current design I think we simply don't allow session level config? 🤔 Or in other words, will we distinguish session variables from system variables? In PG it seems they are also the same set of variables.

I also found interesting notes in tidb doc https://docs.pingcap.com/tidb/stable/system-variables

> Several GLOBAL variables persist to the TiDB cluster. Some variables in this document have a Persists to cluster setting, which can be configured to Yes or No.

xxchan (commented):
The ideas generally LGTM, but needs some clarifications as commented

fuyufjh:
> Regarding the decision where to put an argument: If e.g. the argument relates to a RW instance, how do we decide if we put it into the SQL interface or the web console? Or will we always offer both options?
> 
> Same for process configs: When do we use a CLI arg and when the .toml file?

SQL interface and web console control exactly the same set of parameters. Web console is just an optional GUI.

CLI args is a subset of `.toml` file, containing the frequently-used ones e.g. `--parallelism`

rfcs/0034-system-parameters.md
fuyufjh: I'll improve that part later.

rfcs/0034-system-parameters.md
fuyufjh: Things are easier for single-node systems like PG/MySQL. Here we need to to either 1) move all configs to meta services  2) divided them into 2 sets: local and global. 

I chose 2 because 1 is impossible to avoid. For example, `--memory` must be a local parameter.

rfcs/0034-system-parameters.md
fuyufjh: > I found PG distinguishes SESSION variables and GLOBAL parameters clearly, so I decided not to discuss session variables as well.

Really? I thought PG distinguishes global parameters and session variables strictly.

From my understanding, they are not related for most cases, except one: the GLOBAL parameter decides the default value for new sessions, and a session can SET its parameter to override that default value.

xxchan: > Really? I thought PG distinguishes global parameters and session variables strictly, unlike MySQL/TiDB.

According to [20.1.3. Parameter Interaction via SQL](https://www.postgresql.org/docs/current/config-setting.html#CONFIG-SETTING-SQL-COMMAND-INTERACTION): system level, db level, session level configs seem to be a single set of variables 😇

fuyufjh: Yeah you are right ❤️. So I should discuss the session parameter as well.

rfcs/0034-system-parameters.md
```diff
@@ -0,0 +1,135 @@
...
+
+Currently, we have 3 kinds of configurations:
+
+- Command-line arguments
+- Configuration file `risingwave.toml`
+- Session parameters
+
+These can not meet the requirements for several reasons.
+
+1. **Consistency**. `risingwave.toml` is a local file, which makes it possible that multiple nodes could see different parameters and this is absolutely unacceptable. For example, if some nodes use `data_directory = 'hummock_001'` and others use `data_directory = 'hummock_002'`, the stored data will be corrupted.
```
kwannoel, line 20:
```suggestion
1. **Consistency**. `risingwave.toml` is a local file, which makes it possible that multiple nodes could see different parameters and this is absolutely unacceptable. For example, to configure the remote directory for storing data, users may use the `data_directory` parameter. If some nodes use `data_directory = 'hummock_001'` and others use `data_directory = 'hummock_002'`, the stored data will be corrupted.
```

hzxa21:
related draft PR: https://github.com/risingwavelabs/risingwave/pull/7273

Gun9niR:
> related draft PR: https://github.com/risingwavelabs/risingwave/pull/7273

This PR is drafted before this RPC, so it just includes a very small part of this RPC. Essentially, for demo purpose, it moves one config `state_store` from process config to system parameters, and persists it to etcd. When other nodes join the cluster, it will pull state store URL from meta node. For backward compatibility, meta node will verify `state_store` option from compute node with its own, and reject node join if they differ.

rfcs/0034-system-parameters.md
```diff
@@ -0,0 +1,135 @@
...
+```
+
+```bash
+# Override total_memory_bytes as 2GB
+./risingwave compute-node --total-memory-bytes=2147483648
+```
+
+### Mutability
+
+- Process configs must be **immutable**, because there is no user-interface to change it.
```
kwannoel, line 121: I'm wondering if user has been running some workload for a while, and realize having a lower/higher `parallelism` will better utilize their resources.

In this scenario, `parallelism` is process level config, and is **immutable**.
What should the user do if they want to adjust it, do they have to start a new cluster?
Or is this not a realistic scenario that happens?

rfcs/0034-system-parameters.md
```diff
@@ -0,0 +1,135 @@
...
+    all ==Yes==> system["System Parameters (on Meta)"]
+    process --> cli_args(Command-line arguments)
+    process --> conf_file("Config file (.toml)")
+    system --> sql_cmd(SQL interface: SET / SHOW)
+    system --> cloud(Web Console)
+```
+
+### System Parameters
+
+What we need is a centralized source-of-truth of cluster configurations. Obviously, the best choice for RisingWave is the Meta Service and the underlying etcd.
```
BugenZhao, line 44: I don't think it's a good idea to persist the configuration to the meta store. Here're several reasons:
- Meta store is not designed to be directly manipulated by the user. If one has corrupted some configuration with `SET` and finds that the cluster does not boot anymore, he can do nothing about this.
- Maintaining persisted configuration and consistency with the `clap` definition can be really hard, especially for version compatibility (some famous HTAP database has had a lot of production accidents due to this).
- It seems not convincing to persist configuration like `data_directory`... what if the user really moves the data? Note that the cluster will immediately enter the recovery mode on startup, without giving a chance to the users to set a new `data_directory`.

rfcs/0034-system-parameters.md
fuyufjh: > What should the user do if they want to adjust it, do they have to start a new cluster?

In that case, he/she can restart (or start new, they are equivalent actually) the CN process with new `parallelism` config. No need to restart the cluster.

-------

**Parallelism** is actually a very special case than other parameters. Let me tell the whole story in my mind:

Overall, we have

- Process-level **config** to control the parallelism of this CN node
- System **parameter** `parallelism` as the default parallelism for new streaming jobs. The default values is `0` or `'auto'` which implies the new streaming jobs are supposed to use all available parallel units in all compute nodes.
   - Sure, it can be session-level as well. 

When the system bootup:

1. Compute nodes register itself to Meta node when starting up, including its parallelism, which is controlled by the config `parallelism`
2. Meta service aggregates these numbers to get **total parallelism**, or we called the total number of parallel units. 
   - Note: "Parallel units" are basically one CPU core, but it's logical and is not strictly bound to a CPU core.
3. If the system parameter `parallelism` is set to `0` or `'auto'`, we will use that "total parallelism" in step 2 as the effective value to plan/schedule/run job actors.  Otherwise, the user-set value will be used.

If the cluster changed (e.g. new nodes joined or exiting nodes leaved), it will only affect the "total parallelism" in step 2.

Gun9niR:
For process config, should we have a separate config file (template) for each type of node, or add more sections in `risingwave.toml`, such as:

```toml
[server]
...
[meta]
...
[storage]
...
[compute_node]
# Put the entries from `ComputeNodeOpts` here
[compactor]
# Put the entries from `CompactorOpts` here
```

I was just reviewing the command line arguments for each type of node, and realized they differ a lot. If we put all those arguments into one config file without add more sections, it will be very cluttered, and the subset of configs that will take effect vary from worker type.

fuyufjh:
> > related draft PR: [risingwavelabs/risingwave#7273](https://github.com/risingwavelabs/risingwave/pull/7273)
> 
> This PR is drafted before this RFC, so it just includes a very small part of this RFC. Essentially, for demo purpose, it moves one config `state_store` from process config to system parameters, and persists it to etcd. When other nodes join the cluster, it will pull state store URL from meta node. For backward compatibility, meta node will verify `state_store` option from compute node with its own, and reject node join if they differ.

Such brilliant solution!

rfcs/0034-system-parameters.md
```diff
@@ -0,0 +1,135 @@
...
+References:
+
+- [PostgreSQL: Documentation: 15: ALTER SYSTEM](https://www.postgresql.org/docs/current/sql-altersystem.html)
+- [PostgreSQL: Documentation: 15: 20.1. Setting Parameters](https://www.postgresql.org/docs/current/config-setting.html)
+- [PostgreSQL: Documentation: 15: SHOW](https://www.postgresql.org/docs/current/sql-show.html)
+
+Similar to other system metadata e.g. catalog,
+
+- Changes to global parameters should be propagated to all nodes, in the cluster in a deterministic order ideally
+- When a node reboots or recovers, the latest snapshot of parameters should be passed from Meta Service.
```
BugenZhao, line 68: Could you please elaborate it more about the initialization of the "snapshot" (in meta store)? For example,
- Do we still have configuration file for the cluster? If so, do we only persist the configuration on the first run, or do we also detect changes in the configuration file on restarts?
- Do we persist the default values of the entries that do not appear in `.toml`? We should be cautious about some default values that are "computed" instead of constants, as persisting them may be confusing.

fuyufjh:
> For process config, should we have a separate config file (template) for each type of node, or add more sections in `risingwave.toml`

LGTM for adding more sections. Multiple files may be troublesome for deployment.

rfcs/0034-system-parameters.md
```diff
@@ -0,0 +1,135 @@
...
+./risingwave compute-node --total-memory-bytes=2147483648
+```
+
+### Mutability
+
+- Process configs must be **immutable**, because there is no user-interface to change it.
+- Global configs might be **immutable or mutable**, the mutable ones can be further divided into take effect immediately or need reboot.
+    - **Immutable** e.g. `data_directory` should be rejected upon SET
+    - **Mutable & need reboot** e.g. `heartbeat_interval_ms`
+    - **Mutable & immediately** e.g. `batch_chunk_size`
```
BugenZhao, line 125: I guess there're more things to consider in defining "immediately". As Flink says,

> Because options are read at different point in time when performing operations, it is recommended to set configuration options early after instantiating a table environment.

If we only read the configuration on scheduling (of batch queries...), then it sounds not that "immediate" 😄 but more like a session variable that only affects future queries. If we read it continuously, we may need to consider whether it'll lead to unexpected behavior if the config can be modified at any time.

rfcs/0034-system-parameters.md
fuyufjh: Hmmm. These problems exist indeed, but persisting parameters (instead of losing all changes on reboot) is also a must-have feature for a distribution system.

> Meta store is not designed to be directly manipulated by the user. If one has corrupted some configuration with SET and finds that the cluster does not boot anymore, he can do nothing about this.

We need to equip the operation tool with the ability to fix it. For example, support a new command to `risectl set <parameter> to <value>` to manipulate `etcd` directly.  

> Maintaining persisted configuration and consistency with the `clap` definition can be really hard, especially for version compatibility (some famous HTAP database has had a lot of production accidents due to this).

`clap` definitions must refer to the CLI arguments, right? They are unrelated to system parameters in most cases.

> It seems not convincing to persist configuration like data_directory... what if the user really moves the data? Note that the cluster will immediately enter the recovery mode on startup, without giving a chance to the users to set a new data_directory.

By "persist" I mean to be recoverable after rebooting. It can be mutable, as described in section "mutability".

rfcs/0034-system-parameters.md
fuyufjh: Just to avoid misunderstanding: the "snapshot" here refers to the current system parameters in Meta Service at the beginning of subscription, just like the "snapshot" of the catalog or Hummock versions.

Answers to questions:

> Do we still have configuration file for the cluster? If so, do we only persist the configuration on the first run, or do we also detect changes in the configuration file on restarts?

Basically not. It may be useful for a few cases like https://github.com/risingwavelabs/rfcs/pull/34#issuecomment-1381424149, but generally, I don't think we need such a design.

> Do we persist the default values of the entries that do not appear in `.toml`? We should be cautious about some default values that are "computed" instead of constants, as persisting them may be confusing.

Computed values (like parallelism described in https://github.com/risingwavelabs/rfcs/pull/34#discussion_r1069061515) should be stored as `'auto'` or `0`, so does this solves your concern? In this way, either persisting default values or not can work well, but I still prefer not to persist default values because it allows us to tune the defaults for better performance or whatever.

rfcs/0034-system-parameters.md
BugenZhao: > `clap` definitions must refer to the CLI arguments, right?

Or `serde`? 🥺 If we take a config file as the initialization of these parameters, then we need to have a separate definition of its schema in the code (or we can make it a dynamic `HashMap`).

> It can be mutable, as described in section "mutability".

Yes, but it cannot be mutated without a normal running cluster or `risectl` mentioned above. As the old `data_directory` is invalid, we cannot survive the recovery phase.

> but persisting parameters (instead of losing all changes on reboot) is also a must-have feature for a distribution system.

I'm thinking if we treat all `SET`s to be temporary and just for the users to give it a try, users can always write down the desired config in a new `toml` file and reboot the cluster to apply it. 🤔

rfcs/0034-system-parameters.md
fuyufjh: Exactly. The boundary seems obscure for these cases. I just collectively call them "immediately" to distinguish them from the need-reboot ones.

rfcs/0034-system-parameters.md
BugenZhao: > Basically not. It may be useful for a few cases like [#34 (comment)](https://github.com/risingwavelabs/rfcs/pull/34#issuecomment-1381424149), but generally, I don't think we need such a design.

Does it mean that we need to manually initialize this on the first run, before issuing any streaming jobs? It sounds not that friendly for development and playground purposes.

> but I still prefer not to persist default values

Sounds like a reasonable choice! But we should still be cautious when upgrading if some users rely on the previous default values.

rfcs/0034-system-parameters.md
fuyufjh: > Or `serde`? 🥺 If we take a config file as the initialization of these parameters, then we need to have a separate definition of its schema in the code (or we can make it a dynamic HashMap).

In my mind, the system parameters are all simple strings or integers so se/deserialization won't be a big issue.

> Yes, but it cannot be mutated without a normal running cluster or risectl mentioned above. As the old data_directory is invalid, we cannot survive the recovery phase.

Perhaps it also needs that operational tool: "We need to equip the operation tool with the ability to fix it" (mentioned in previous comment)

Anyway, `data_directory` is very special in many aspects: It's required to be set; it can hardly be changed when RisingWave is running; it need to be mutated offline. Things may get a little dirty for it.

> I'm thinking if we treat all SETs to be temporary and just for the users to give it a try, users can always write down the desired config in a new toml file and reboot the cluster to apply it. 🤔 I cannot remember well but I guess this is also the Flink way?

Hmmm. Interesting. It's indeed a popular way in big data systems. 

That's basically a centralized but non-persistent config store. It makes senses, but since we had a metadata database, why not just persist it? I think in most cases this looks better (except `data_dir`, I know 😇)

Gun9niR:
> > For process config, should we have a separate config file (template) for each type of node, or add more sections in `risingwave.toml`
> 
> LGTM for adding more sections. Multiple files may be troublesome for deployment.

Is it a good idea to put configs that are bound to be different among nodes like host address in the same config file 🤔 ?

rfcs/0034-system-parameters.md
fuyufjh: > Does it mean that we need to manually initialize this on the first run, before issuing any streaming jobs? It sounds not that friendly for development and playground purposes.

IIUC, we use minio for development and playground purposes now, and that's why we can use a default value. We can keep this behavior unchanged. 

But for productional deployment or cloud platform, a unique S3 bucket must be manually specified for sure.

fuyufjh:
> > > For process config, should we have a separate config file (template) for each type of node, or add more sections in `risingwave.toml`
> > 
> > 
> > LGTM for adding more sections. Multiple files may be troublesome for deployment.
> 
> Is it a good idea to put configs that are bound to be different among nodes like host address in the same config file 🤔 ?

Then use CLI args instead, if it makes you feel better 🤣

rfcs/0034-system-parameters.md
BugenZhao: This acts kind of like a mobile App and we need to provide several mechanisms to make it work, like entering the "emergency mode" and providing fixing utilities if the cluster frequently fails, or guiding them to contact customer services for asking help. 😄

I do agree we should clarify the difference between "process" configs and "cluster" configs, but I'm still doubting the possible issues and complexities for persisting it. I feel like we are taking responsibility for ourselves that could otherwise belong to the user or DevOps. 😂

rfcs/0034-system-parameters.md
BugenZhao: Not only the `data_directory`, but also the `state_store_url` which tells which state store backend we should use. 🤔

Gun9niR:
> > > > For process config, should we have a separate config file (template) for each type of node, or add more sections in `risingwave.toml`
> > > 
> > > 
> > > LGTM for adding more sections. Multiple files may be troublesome for deployment.
> > 
> > 
> > Is it a good idea to put configs that are bound to be different among nodes like host address in the same config file 🤔 ?
> 
> Then use CLI args instead, if it makes you feel better 🤣

I agree CLI args is more suitable. So these configs shouldn't be written in the **template** config file, although the users can surely add them in their own files, right? Because there's no appropriate default value.

rfcs/0034-system-parameters.md
fuyufjh: The root problem is that our system is distributed, even the Meta Service. Therefore, when someone needs to change a system parameter, she/he must change it in multiple containers/machines. That would be a pain for both human and cloud platform, and ...... looks like a design mistake 🤪  

I admit that Flink insists to use local files and I have no idea for the reasons. 🤣

fuyufjh: I suddenly remembered that K8s allows to mount a config as a file: https://kubernetes.io/docs/concepts/configuration/configmap/

This looks a graceful way! Although somewhat coupled with K8s

rfcs/0034-system-parameters.md
jon-chuang: Basically, is it correct to say that `stream_parallelism` is only a meaningful config for dev/on-prem, to set a stream-job level default, but those using cloud should never have to care about it due to auto-redistribution?

Btw, I’m quite unclear, but do we currently have an API for scaling individual fragments? Is this API an SQL API or is it a HTTP/gRPC endpoint to meta?

jon-chuang: Also, is there a way for the user to hint the starting parallelism? For instance, even if cloud may auto distribute, it is unclear if it is better to start with small parallelism or high for a given stream job. User may know better.

So something like `create materialized view (stream_job_parallelism_hint=3) …` when the user thinks the stream job requires small throughput.

fuyufjh: > Btw, I’m quite unclear, but do we currently have an API for scaling individual fragments

Yes, See the design doc at https://singularity-data.quip.com/RubDA4a51oS0/Low-level-Scaling-Interface-on-Meta-Service

> So something like `create materialized view (stream_job_parallelism_hint=3)` … when the user thinks the stream job requires small throughput.

That's exactly the session variable `streaming_parallelism` does. See  https://github.com/risingwavelabs/risingwave/pull/7370. 
> but to me it is unclear what this variable does if we are going to auto scale, whereas a hint is at least clearer in that the parallelism may be modified by the system or by external operator.

I understand this seems to be somewhat inconsistent. In the current design, `streaming_parallelism` is just a hint for the subsequent `CREATE MVIEW` command. The syntax you proposed LGTM as well. After the MV is created, the only way to manipulate its parallelism is via the Scaling API (the [doc](https://singularity-data.quip.com/RubDA4a51oS0/Low-level-Scaling-Interface-on-Meta-Service) above). 

The reason is that we consider scaling as an advanced operational feature that **should be part of the cloud platform** instead of the database kernel, especially for the **policy** i.e. how and when to do scaling. Hence, we divide the feature "scaling" into 2 parts: 1) the cloud platform make decisions about how & when, and then 2) the kernel simply **executes** these directives. That's the design philosophy behind [that API](https://singularity-data.quip.com/RubDA4a51oS0/Low-level-Scaling-Interface-on-Meta-Service). However, the only exception is creating an MView, during which the meta must decide the parallelism without cloud platform engaged. In that case, I think the best we can do is to use `stream_parallelism` for each fragment.

rfcs/0034-system-parameters.md
CAJan93: IMHO ConfigMaps may actually be a good solution here. I am not sure how we would make this work in the risedev environment. Btw: Under the hood CMs, like all K8s objects, are also stored in ETCD. 

> Meta store is not designed to be directly manipulated by the user. If one has corrupted some configuration with SET and finds that the cluster does not boot anymore, he can do nothing about this. 

Do you think it would  help to version the configurations? Maybe this would make it simpler to fall back to a previous version if an update to the settings broke something. ETCD supports a `--prefix` operation, so you could retrieve all values belonging to a setting independent from the timestamp like this.  

```bash 
etcdctl put some_setting_23_01_27_13_22 val1 # original value
etcdctl put some_setting_23_01_28_09_00 val1 # updated value
etcdctl get some_setting --prefix # retrieve all values
some_setting_23_01_27_13_22 val1 
some_setting_23_01_28_09_00 val1
```

xxchan: > Do you think it would help to version the configurations

It's worth a separate RFC 🫣

CAJan93: Sounds good. But I believe first we would have to decide on how to store the configs, before deciding if/how we want to version them.

BugenZhao: +1 If we really want to persist them to the meta store, we should carefully think of a way to store them. For example, we should introduce something like the "field number" in protobuf, so that we can handle the backward compatibility more easily.

arkbriar: Well, I totally agree with the centralized system-wide parameters, no matter where they are stored, for the reason that it's painful for the operator to dispatch those global vars like `data_directory` now 👍 . IMO, it can be stored in 
- the etcd as a part of the metadata because we may want to change them afterward
- a local and consistent file mounted on the meta nodes so that the meta node can be the single source of truth, as all of the other components will contact meta before they get started

rfcs/0034-system-parameters.md
```diff
@@ -0,0 +1,135 @@
...
+
+References:
+
+- [PostgreSQL: Documentation: 15: ALTER SYSTEM](https://www.postgresql.org/docs/current/sql-altersystem.html)
+- [PostgreSQL: Documentation: 15: 20.1. Setting Parameters](https://www.postgresql.org/docs/current/config-setting.html)
+- [PostgreSQL: Documentation: 15: SHOW](https://www.postgresql.org/docs/current/sql-show.html)
+
+Similar to other system metadata e.g. catalog,
+
+- Changes to global parameters should be propagated to all nodes, in the cluster in a deterministic order ideally
```
arkbriar, line 67: Considering the nature of distributed systems, there will be cases where multiple values of the same parameter exist on various nodes, no matter how we optimize the notification mechanism. Will we introduce things like what is described in the Online Schema Change paper to ensure consistency and atomicity?

rfcs/0034-system-parameters.md
arkbriar: I don't think the `data_directory` will be a big problem as long as it's not mutable and the deployment tool enforces the value. What really matters is how we keep the caches on various nodes updated and how to make sure things do not break when parameter change happens.

rfcs/0034-system-parameters.md
```diff
@@ -0,0 +1,135 @@
...
+
+- [PostgreSQL: Documentation: 15: 20.1. Setting Parameters](https://www.postgresql.org/docs/current/config-setting.html)
+
+Previously, we had already introduced [session parameters in RisingWave](https://github.com/risingwavelabs/risingwave/blob/53f7e0db772ac7e51773791bb8301624ed763ae8/src/common/src/session_config/mod.rs#L265). Some of them should be system parameters like `batch_enable_sort_agg` or `query_mode`. Postgres **enforces** that session parameters must also be system paramters, we can follow this rule as well.
+
+### Process Configs
+
+Let's rethink the current design of CLI options and config files. ([#5676](https://github.com/risingwavelabs/risingwave/issues/5676)) Given the existence of system parameters, some existing parameters may need to be moved to system parameters, like `data_directory` as mentioned before.
+
+We propose to consider CLI options as a way to override config files (i.e. in term of priority, CLI options -> config file -> defaults); as a result, the items in CLI options are a subset of the config file. The reasons for this include
```
arkbriar, line 90: Shall we also consider supporting env vars as one of the config sources? In containerized deployments, it is way easier to set an env var than update a mounted file, not to mention that RisingWave is a distributed system that makes config dispatching a mess: it's hard to maintain a lot of local config files.

rfcs/0034-system-parameters.md
arkbriar: Can we say that local or process configs are immutable during runtime to avoid confusion? Because users can always kill and restart a process with different config values. 

> because there is no user-interface to change it.

That won't be true if the cloud portal supports updating with a reboot.

rfcs/0034-system-parameters.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Style
+
+- Process configs follow the style of TOML (lower-case, underscore) and ordinary CLI arguments (lower-case, hyphen)
+- Global configs should be flatten unlike TOML. Consider add a prefix e.g. `stream_foo`, `batch_foo` if it seems ambiguous.
+- Dangerous configs should be named with keyword `developer` or `unsafe` as a warning.
+
+## Discussions
+
+1. Do we need database parameters, global (instance-wide) parameters, or both?
```
arkbriar, line 135: 🤔 Are there real cases or just a proposal? Adding levels will make the configuration complicated. I'd say that instance-wide and object(table, source, mv)-specific parameters should be enough.

rfcs/0034-system-parameters.md
CAJan93: > [...] not to mention that RisingWave is a distributed system that makes config dispatching a mess: it's hard to maintain a lot of local config files.

IMHO K8s ConfigMaps may help here, since dispatch and updates will happened automatically. Maybe it would be a good alternative to combine CMs with env vars? A local env var could overwrite the CM and you could thus customise the global configs for every pod.

rfcs/0034-system-parameters.md
CAJan93: Do you mean [Online, Asynchronous Schema Change in F1](https://storage.googleapis.com/pub-tools-public-publication-data/pdf/41376.pdf)  by Rae et al.? As I understand [this summary](https://pages.cs.wisc.edu/~zuyu/summaries/cs764/f1Schema) they basically use 2PC to change schemas, right? 

Maybe we could still rely on ETCD (with or without abstracting via a K8s CM), since it offers strong consistency? A 2PC using ETCD could look like this maybe: 

1. Client issues settings change against system 
2. ETCD changes configs 
3. Pods are alerted about config change (if we use CM, K8s would do that for us)
4. Meta (or some other components) asks each pod if it is ready to apply the changes. If all reply yes, then we apply. 

Just a suggestion :)

rfcs/0034-system-parameters.md
xxchan: In PG there's only a set of variables, and can be set at different levels (session, role, database, instance).

> object(table, source, mv)-specific parameters

But this seems similar to database parameters. IMHO we didn't consider such things in the RFC? 🤔

rfcs/0034-system-parameters.md
arkbriar: Yep, it's the exact paper that I referred.

>Maybe we coul rely on ETCD (with or without abstracting via a K8s CM), since it offers strong consistency?

🤔 Sounds like a feasible solution.

rfcs/0034-system-parameters.md
arkbriar: > But this seems similar to database parameters. IMHO we didn't consider such things in the RFC? 🤔

You're right. Let's be PG-similar.

fuyufjh: Not decided yet. But anyway we can add support for database parameters after the completion of global parameters.

fuyufjh:
Shall we store the parameters in etcd in a simple readable manner? Such as plain key-values (in human-readable strings e.g. `"114514"`, `"true"`), so that we (the operation person or cloud platform) can easily initialize or modify data directly, even without Meta Service.

In this way, the global parameters can be modified more easily, which is one of the benefits of using a config file, while also get consistency and replicated among multiple nodes. 

cc. @Gun9niR @BugenZhao

hzxa21:
> Shall we store the parameters in etcd in a simple readable manner? Such as plain key-values (in human-readable strings e.g. `"114514"`, `"true"`), so that we (the operation person or cloud platform) can easily initialize or modify data directly, even without Meta Service.
> 
> In this way, the global parameters can be modified more easily, which is one of the benefits of using a config file, while also get consistency and replicated among multiple nodes.

+1

rfcs/0034-system-parameters.md
```diff
@@ -0,0 +1,135 @@
...
+
+```sql
+SET parameter TO value / 'value' / DEFAULT;
+```
+
+References:
+
+- [PostgreSQL: Documentation: 15: 20.1. Setting Parameters](https://www.postgresql.org/docs/current/config-setting.html)
+
+Previously, we had already introduced [session parameters in RisingWave](https://github.com/risingwavelabs/risingwave/blob/53f7e0db772ac7e51773791bb8301624ed763ae8/src/common/src/session_config/mod.rs#L265). Some of them should be system parameters like `batch_enable_sort_agg` or `query_mode`. Postgres **enforces** that session parameters must also be system paramters, we can follow this rule as well.
```
Gun9niR, line 84: Session parameters seems to conflict with the definition of system parameters, since the values can vary from session to session (on different fe nodes). So I think session parameters should be out of scope for this RFC.

Gun9niR (approved):
🚀

st1page:
feedback: 
In recent POC, the user tried to upgrade a RisingWave cluster with the too-old version and a system parameter become incompatible. So we have to alter the system parameter in the etcd manually. 

This really helps
> Shall we store the parameters in etcd in a simple readable manner? Such as plain key-values (in human-readable strings e.g. "114514", "true"), so that we (the operation person or cloud platform) can easily initialize or modify data directly, even without Meta Service.

And I guess we need this for Ops.
> > Meta store is not designed to be directly manipulated by the user. If one has corrupted some configuration with SET and finds > that the cluster does not boot anymore, he can do nothing about this.
>
>We need to equip the operation tool with the ability to fix it. For example, support a new command to risectl set <parameter> to <value> to manipulate etcd directly.

PR #35, opened 2023-01-14
RFC: a detailed design for create connection

tabVersion:
preview: https://github.com/risingwavelabs/rfcs/blob/tab/create-connection/rfcs/0035-create-connection.md

### DIFF
--- a/rfcs/0035-create-connection.md
+++ b/rfcs/0035-create-connection.md
@@ -18,19 +18,22 @@ A connection is used to describe how to connect to an external system that users
 ```sql
 -- create connection
 CREATE CONNECTION (IF NOT EXIST) {{ connection_name }}
-  TO {{ connection_type }}
-  ( {{ field }} = {{ value }}, ...);
+  with ( 
+    connection_type = '{{ connection_type }}',
+    {{ field }} = {{ value }}, ...);
 
 -- create source with connection
 CREATE SOURCE {{ source_name }} ( {{ field }} = {{ value }}, ... )
-  FROM CONNECTION {{ connection_name }}
-  with ( {{ field }} = {{ value }}, ...);
+  with ( 
+    connection = '{{ connection_name }}',
+    {{ field }} = {{ value }}, ...);
 
 -- create sink with connection
 CREATE SINK {{ sink_name }} ( {{ field }} = {{ value }}, ... )
   FROM {{ source_name }}
-  INTO CONNECTION {{ connection_name }}
-  with ( {{ field }} = {{ value }}, ...);
+  with ( 
+    connection = '{{ connection_name }}',
+    {{ field }} = {{ value }}, ...);
 ```
 
 If any field is specified both in `with` clause and `CREATE CONNECTION` statement, the value in `with` clause will be used.
@@ -40,40 +43,37 @@ If any field is specified both in `with` clause and `CREATE CONNECTION` statemen
 An AWS Private Link establishes a private connection between a VPC and a service hosted on AWS. It is a secure and scalable way to connect to AWS services from within a VPC. The following fields are required to create an AWS Private Link connection:
 
 ```sql
-CREATE CONNECTION demo_connection TO AWS_PRIVATE_LINK
-  (SERVICE NAME = 'com.amazonaws.vpce.us-east-1.vpce-svc-xxxxxxxxxxxxxxxxx',
-  AVAILABILITY ZONES = array('us-east-1a', 'us-east-1b', 'us-east-1c');
+CREATE CONNECTION demo_connection
+  with (
+    connection_type = 'aws_private_link',
+    service_name = 'com.amazonaws.vpce.us-east-1.vpce-svc-xxxxxxxxxxxxxxxxx',
+    availability_zones = 'us-east-1a,us-east-1b,us-east-1c');
 ```
 
 |Field|Value|Required|Description|
 |--|--|--|--|
 |`SERVICE NAME`| varchar | yes | The name of the AWS PrivateLink service. |
-|`AVAILABILITY ZONES`| varchar[] | yes | The AWS availability zones in which the service is accessible. |
+|`AVAILABILITY ZONES`| varchar | yes | The AWS availability zones in which the service is accessible. Listing the zones in a string, separated by comma. |
 
 > **Notice**:
 >
 > * After creating a connection, users must manually approve the connection from a Risingwave VPC in AWS console.
 > * When creating a private link to AWS MSK, users must fill in all availability zones of the MSK cluster.
 
-##### Internal table for AWS Private Link
-
-The `rw_aws_private_link` table stores the information of the AWS Private Link connection. The following fields are stored in the table:
-
-|Field|Value|Description|
-|--|--|--|
-|`id`| varchar | The name of the connection. |
-|`principle`| varchar | The AWS Principal that Risingwave will use to connect to the VPC endpoint. |
-
 #### Kafka
 
 A Kafka connection establishes a link to a Kafka cluster. Users can use Kafka connections to create Kafka sources and sinks. The following fields are required to create a Kafka connection:
 
 ```sql
-CREATE CONNECTION demo_connection_1 TO KAFKA
-  (PROPERTIES.BOOTSTRAP.SERVER = 'broker1:9092,broker2:9092');
+CREATE CONNECTION demo_connection_1
+  with (
+    connection_type = 'kafka',
+    PROPERTIES.BOOTSTRAP.SERVER = 'broker1:9092,broker2:9092');
 
-CREATE CONNECTION demo_connection_2 TO KAFKA
-  (PROPERTIES.BOOTSTRAP.SERVER.PRIVATE_LINK = ['broker1:9092' USING AWS_PRIVATE_LINK {{ connection_name }} (PORT 9092), 'broker2:9092' USING AWS_PRIVATE_LINK {{ connection_name }} (PORT 9092)]];
+CREATE CONNECTION demo_connection_2
+  with (
+    connection_type = 'kafka',
+    PROPERTIES.BOOTSTRAP.SERVER.PRIVATE_LINK = ['broker1:9092' USING AWS_PRIVATE_LINK {{ connection_name }} (PORT 9092), 'broker2:9092' USING AWS_PRIVATE_LINK {{ connection_name }} (PORT 9092)]];
 ```
 
 For general usage, the following fields are allowed:
@@ -98,11 +98,13 @@ For security related fields (SSL/SASL), the fields are identical to [this doc](h
 A Kinesis connection establishes a link to a Kinesis stream. Users can use Kinesis connections to create Kinesis sources and sinks. The following fields are required to create a Kinesis connection:
 
 ```sql
-CREATE CONNECTION demo_connection TO KINESIS
-  (aws.region='user_test_topic',
-   endpoint='172.10.1.1:9090,172.10.1.2:9090',
-   aws.credentials.role.arn='arn:aws-cn:iam::602389639824:role/demo_role',
-   aws.credentials.role.external_id='demo_external_id',
+CREATE CONNECTION demo_connection
+  with (
+    connection_type = 'kinesis',
+    aws.region='user_test_topic',
+    endpoint='172.10.1.1:9090,172.10.1.2:9090',
+    aws.credentials.role.arn='arn:aws-cn:iam::602389639824:role/demo_role',
+    aws.credentials.role.external_id='demo_external_id',
    ...
   );
 ```
@@ -124,16 +126,18 @@ Accepted fields are listed below, which is a subset of [the doc](https://www.ris
 #### Pulsar
 
 ```sql
-CREATE CONNECTION demo_connection TO PULSAR
-  (service.url = 'pulsar://localhost:6650/',
-  admin.url='http://localhost:8080',);
+CREATE CONNECTION demo_connection
+  with (
+    connection_type = 'pulsar',
+    service.url = 'pulsar://localhost:6650/',
+    admin.url='http://localhost:8080',);
 ```
 
 The accepted fields are listed below, which is a subset of [the doc](https://www.risingwave.dev/docs/current/create-source-pulsar/#parameters).
 
 |Field|Required|Description|
 |--|--|--|
-|`service.url`| Required| Address of the Pulsar service.	|
+|`service.url`| Required| Address of the Pulsar service.|
 |`admin.url`|Required| Address of the Pulsar admin.|
 
 **Notice**: Risingwave will not check the connection to Pulsar is valid or not.
--- a/rfcs/0035-create-connection.md
+++ b/rfcs/0035-create-connection.md
@@ -16,10 +16,24 @@ A connection is used to describe how to connect to an external system that users
 ### Syntax
 
 ```sql
+-- create connection
 CREATE CONNECTION (IF NOT EXIST) {{ connection_name }}
   TO {{ connection_type }}
   ( {{ field }} = {{ value }}, ...);
+
+-- create source with connection
+CREATE SOURCE {{ source_name }} ( {{ field }} = {{ value }}, ... )
+  FROM CONNECTION {{ connection_name }}
+  with ( {{ field }} = {{ value }}, ...);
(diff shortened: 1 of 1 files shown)


rfcs/0035-create-connection.md
```diff
@@ -0,0 +1,143 @@
...
+  FROM {{ source_name }}
+  INTO CONNECTION {{ connection_name }}
+  with ( {{ field }} = {{ value }}, ...);
+```
+
+If any field is specified both in `with` clause and `CREATE CONNECTION` statement, the value in `with` clause will be used.
+
+#### AWS Private Link
+
+An AWS Private Link establishes a private connection between a VPC and a service hosted on AWS. It is a secure and scalable way to connect to AWS services from within a VPC. The following fields are required to create an AWS Private Link connection:
```
fuyufjh, line 40: IIUC, Private Link is orthogonal with the service to connect, and it only makes sense when being used with the definition of source/sink together, right?

PS. I guess you borrowed the idea from https://materialize.com/docs/sql/create-connection/

rfcs/0035-create-connection.md
```diff
@@ -0,0 +1,143 @@
...
+|`AVAILABILITY ZONES`| varchar[] | yes | The AWS availability zones in which the service is accessible. |
+
+> **Notice**:
+>
+> * After creating a connection, users must manually approve the connection from a Risingwave VPC in AWS console.
+> * When creating a private link to AWS MSK, users must fill in all availability zones of the MSK cluster.
+
+##### Internal table for AWS Private Link
+
+The `rw_aws_private_link` table stores the information of the AWS Private Link connection. The following fields are stored in the table:
```
fuyufjh, line 60: Metadata is stored in etcd with protobuf encoding, rather than relational tables.

rfcs/0035-create-connection.md
```diff
@@ -0,0 +1,143 @@
...
+
+```sql
+-- create connection
+CREATE CONNECTION (IF NOT EXIST) {{ connection_name }}
+  TO {{ connection_type }}
+  ( {{ field }} = {{ value }}, ...);
+
+-- create source with connection
+CREATE SOURCE {{ source_name }} ( {{ field }} = {{ value }}, ... )
+  FROM CONNECTION {{ connection_name }}
```
fuyufjh, line 26: ’With connection’ may be better than ‘from/to connection’ because: 

1. It’s consistent between source & sink
2. To avoid the potential ambiguity of the word ‘from’

rfcs/0035-create-connection.md
```diff
@@ -0,0 +1,143 @@
...
+|Field|Required|Description|
+|--|--|--|
+|`service.url`| Required| Address of the Pulsar service.	|
+|`admin.url`|Required| Address of the Pulsar admin.|
+
+**Notice**: Risingwave will not check the connection to Pulsar is valid or not.
+
+## Future possibilities
+
+* store some sensitive information, eg. passwords and SSL keys, in a encrypted way (use `CREATE SECRET`)
```
fuyufjh, line 143: Question: In Flink if you connect to a catalog e.g. [JDBC database](https://nightlies.apache.org/flink/flink-docs-master/docs/connectors/table/jdbc/#jdbc-catalog), you also automatically get the schema of tables in that database. Will we support this? Does Materialize support this?

rfcs/0035-create-connection.md
```diff
@@ -0,0 +1,143 @@
...
+authors:
+  - "Bohan Zhang"
+start_date: "2023/01/15"
+---
+
+# RFC: Create Connection
+
+## Motivation
+
+A connection is used to describe how to connect to an external system that users want to read data from. Once created, a connection is reusable across `CREATE SOURCE` and `CREATE SINK` statements. This RFC proposes a new `CREATE CONNECTION` statement to create a connection.
```
fuyufjh, line 12: Any stronger motivation? I mean, like, beside the convenience it brings, is there anything that was impossible but become doable after it?

rfcs/0035-create-connection.md
```diff
@@ -0,0 +1,143 @@
...
+|--|--|--|
+|`aws.region`| yes | AWS service region. For example, US East (N. Virginia). |
+|`endpoint`| Optional |URL of the entry point for the AWS Kinesis service.|
+|`aws.credentials.access_key_id`| Conditional | This field indicates the access key ID of AWS. It must appear in pairs with aws.credentials.secret_access_key.|
+|`aws.credentials.secret_access_key`|Conditional | This field indicates the secret access key of AWS. It must appear in pairs with aws.credentials.access_key_id.|
+|`aws.credentials.session_token`|Optional| The session token associated with the temporary security credentials.|
+|`aws.credentials.role.arn`|Optional| The Amazon Resource Name (ARN) of the role to assume.|
+|`aws.credentials.role.external_id`|Optional|The [external id](https://aws.amazon.com/blogs/security/how-to-use-external-id-when-granting-access-to-your-aws-resources/) used to authorize access to third-party resources.|
+
+**Notice**: Risingwave will not check the connection to Kinesis is valid or not.
```
fuyufjh, line 122: Reason?

rfcs/0035-create-connection.md
tabVersion: yes, private link provides a way to allow the service in our vpc to access resources in another vpc but it cannot be reversed.

rfcs/0035-create-connection.md
tabVersion: then implementing `show connections` should be enough

rfcs/0035-create-connection.md
tabVersion: there is no proper abstraction other than `connection` to connect private link with the existing reader 🤣 
@wyhyhyhyh has asked for this feature in the last Q

rfcs/0035-create-connection.md
tabVersion: in `create connection`, we don't specify topic/stream. We may don't have access to list topics globally and there is no proper API for the validation.

hzxa21 (commented):
I am also a little bit skeptical about the benefits brought by this RFC. IMO, the benefits can be:
1. It provides a way to trigger the creation and approval process for AWS private link.
2. It makes it easier to reuse create mulitple sources/sinks on the same external system since user just needs to fill-in the cluster information once.

Several questions just come from the head of my mind when reading this RFC:
- As mentioned in this RFC, the behaviors are different for different types of connection. For example, we do trigger the approval process and validate the connection for AWS Private Link while for other connections, we just store the information without validation. This might complicate things and makes me think this feature is more specific to AWS Private Link, not other systems.
- If this feature is more specific to AWS Private Link, it looks like a cloud feature instead of a kernel/SQL feature. Have we considered triggering the creation and approval process in the cloud service/portal instead of via SQL? I am not sure whether this is doable but this can simplify the kernel implementation as well as the SQL syntax. 
- Will we still support the current CREATE SOURCE/SINK syntax, in which we provide the cluster information in the WITH clause? In other words, can user still create source/sink without create connection first? If yes, it may be a burden to implement a new source/sink. If not, user may get confused.

StrikeW:
> * If this feature is more specific to AWS Private Link, it looks like a cloud feature instead of a kernel/SQL feature. Have we considered triggering the creation and approval process in the cloud service/portal instead of via SQL? I am not sure whether this is doable but this can simplify the kernel implementation as well as the SQL syntax.

Here is the background to implement private link support on the kernel side. https://www.notion.so/risingwave-labs/RFC-Using-AWS-PrivateLink-to-Connect-a-Kafka-Instance-7b33defa8af14caab4122fb6f06a5cb9
cc @mikechesterwang @Nebulazhang

tabVersion:
> As mentioned in this RFC, the behaviors are different for different types of connection. For example, we do trigger the approval process and validate the connection for AWS Private Link while for other connections, we just store the information without validation. This might complicate things and makes me think this feature is more specific to AWS Private Link, not other systems.

I want to validate connection in this step but we may not find an API for only validating brokers without a specific topic. We can require a topic here for validation. 

> Have we considered triggering the creation and approval process in the cloud service/portal instead of via SQL? 

The user's kafka cluster is deployed in different vpc's and the client side must connect to the broker's leader node. If the client's request address is wrong, then the linked broker will return the correct broker's address within its vpc. This requires that source/sink can connect to all brokers and can rewrite the broker's address. 
We would like to make this feature available to open-source users rather than directing them to connect via vpc peering

> Will we still support the current CREATE SOURCE/SINK syntax, in which we provide the cluster information in the WITH clause? 

Yes, users can still create source/sink without creating connection. If they provides both, options in with clause will be used. 

> If yes, it may be a burden to implement a new source/sink.

For kafka, yes, we may support private link for it. But for others, no, we only treat them as a hashmap and try to get lacking fields.

rfcs/0035-create-connection.md
fuyufjh: Emmm, they ("private link" and "connection") look independent with each other. "Connection" here provides not only the network access to VPC but also the very credentials it needs to connect to a specific data source/sink.

I mean, since you are proposing to allow users to write AWS VPC credentials along with the Kafka/Kinis properties e.g.

```sql
CREATE CONNECTION demo_connection
  with (
    connection_type = 'kinesis',
    aws.region='user_test_topic',
    endpoint='172.10.1.1:9090,172.10.1.2:9090',
    aws.credentials.role.arn='arn:aws-cn:iam::602389639824:role/demo_role',
    aws.credentials.role.external_id='demo_external_id',
   ...
  )
```

Then, why not just allow users to write AWS VPC credentials within `CREATE SOURCE`:

```sql
CREATE SOURCE demo_source
  with (
    connector_type = 'kinesis',
    aws.region='user_test_topic',
    endpoint='172.10.1.1:9090,172.10.1.2:9090',
    aws.credentials.role.arn='arn:aws-cn:iam::602389639824:role/demo_role',
    aws.credentials.role.external_id='demo_external_id',
   ...
  )
```

Disclaimer: I am not arguing this syntax is better. I am saying the proposed solution seems not to be an exact answer to the problem/motivation.

rfcs/0035-create-connection.md
fuyufjh: e.g.

```sql
CREATE CATALOG my_catalog WITH(
    'type' = 'jdbc',
    'default-database' = '...',
    'username' = '...',
    'password' = '...',
    'base-url' = '...'
);

USE CATALOG my_catalog;

SHOW TABLES;
-- Will print a list of tables

DESCRIBE TABLE foo;
-- Will print the schema of that table.
```

fuyufjh:
> > If yes, it may be a burden to implement a new source/sink.

> For kafka, yes, we may support private link for it. But for others, no, we only treat them as a hashmap and try to get lacking fields.

I guess Patrick's idea is to add these `aws....` fields into other connectors' properties like Kinesis (example in https://github.com/risingwavelabs/rfcs/pull/35#discussion_r1087470892). In this way, they will also support AWS Private Link, right?

StrikeW:
> > > If yes, it may be a burden to implement a new source/sink.
> > 
> > 
> > For kafka, yes, we may support private link for it. But for others, no, we only treat them as a hashmap and try to get lacking fields.
> 
> I guess Patrick's idea is to add these `aws....` fields into other connectors' properties like Kinesis (example in [#35 (comment)](https://github.com/risingwavelabs/rfcs/pull/35#discussion_r1087470892)). In this way, they will also support AWS Private Link, right?

IMO, the RFC introduces a new concept `Connection` to users. But the goals it wants to achieve (e.g. private link and reusable across `CREATE SOURCE` and `CREATE SINK` statements) can also be provided by the existing `CREATE SOURCE/CREATE SINK` statement (e.g. specify those `aws.xx` fields in the `WITH` clause). So I think it is important to clarify the motivation to introduce `Connection`. cc @neverchanje

mikechesterwang:
> Here is the background to implement private link support on the kernel side. https://www.notion.so/risingwave-labs/RFC-Using-AWS-PrivateLink-to-Connect-a-Kafka-Instance-7b33defa8af14caab4122fb6f06a5cb9
cc @mikechesterwang @Nebulazhang

My idea is to hide all these details from users. They can add private links through the cloud portal. Kernel can just watch to a config file which contains all dns resolve mapping results.

rfcs/0035-create-connection.md
tabVersion: I don't think it is related to `connection`, which specifies the parameters used when connecting to external systems. 
The feature described above needs further investigation.

tabVersion:
Connection should be reusable, private link is only one of its functions and connection provides the basis for creating source/sink with a secret key later. Currently, the connection is not checked because we don't find a suitable api for it, we may fix it in future designs.
Also this level of abstraction is necessary, specifying the private link in the with clause is too complicated and needs to be matched with the kafka broker address one by one. But this can be left for later discussion.

The core of the current effort is to support private links, and in the short term I support providing this capability by 
```sql
create aws-private-link {{ name }} with (
  SERVICE NAME = 'service_name',
  AVAILABILITY ZONES = 'az1,az2'
);

create source/sink {{ name }} ... with (
  connector = 'kafka',
  properties.bootstrap.server = 'ip1:port1,ip2:port2',
  private.links = 'link1:port1,link2:port2',
  ...
)
```

The problem is that CN does not have private link specifications, which requires an rpc back to meta.

rfcs/0035-create-connection.md
tabVersion: adding a default topic to do the validation is acceptable but has concerns about misuse.

hzxa21:
> > > If yes, it may be a burden to implement a new source/sink.
> > 
> > 
> > For kafka, yes, we may support private link for it. But for others, no, we only treat them as a hashmap and try to get lacking fields.
> 
> I guess Patrick's idea is to add these `aws....` fields into other connectors' properties like Kinesis (example in [#35 (comment)](https://github.com/risingwavelabs/rfcs/pull/35#discussion_r1087470892)). In this way, they will also support AWS Private Link, right?

Exactly. I am a littble bit conservative about introducing new SQL concept to kernel user since now we need to teach them what `CONNECTION` is, explain its relationship with `SOURCE`/`SINK`, and instruct them `SOURCE`/`SINK` can be created w/ or w/o `CONNECTION`. For cloud user, it may be fine because things can be hidden underneath.

If we do think introducing `CONNECTION` is necessary, I suggest we disallow `CREATE SOURCE/SINK` without `CREATE CONNECTION` to make the semantics clearer.

tabVersion:
new connection syntax:

```sql
create connection connection_name with (
    type = 'private_link',
    provider = 'aws',
    service.name = 'xxx',
    availability.zones = '[{["a", "b", "c"]", "port": 8080}, {...}, {...}]'
);
```

StrikeW:
> new connection syntax:
> 
> ```sql
> create connection connection_name with (
>     type = 'private_link',
>     provider = 'aws',
>     service.name = 'xxx',
>     availability.zones = '[{"az": ["a", "b", "c"], "port": 8080}, {...}, {...}]'
> );
> ```
>
>```sql
> create source/sink {{ name }} ... with (
>   connector = 'kafka',
>   properties.bootstrap.server = 'ip1:port1,ip2:port2',
>   private.links = 'connection_name',
>   ...
> );
>```
> we force users to create connection before creating a source using private link.

After retrospected the concepts of AWS privatelink, I think the above syntax can be improved.
The private link connection to be created is a vpc endpoint to access the endpoint service provided by the user. We only need a service name and AZs to create the endpoint, the ports are the listening ports of the endpoint service which can be changed. The vpc endpoint doesn't need to aware those ports, since we will use DNS name to access the endpoint service.

So I suggest putting the target ports to the CREATE SOURCE statement, and make it explicitly to show the mapping between the source broker address and the target AZ and port.
```sql
create connection connection_name with (
    type = 'private_link',
    provider = 'aws',
    service.name = 'xxx',
    availability.zones = '["az1", "az2", "az3"]'
);

create source/sink {{ name }} ... with (
  connector = 'kafka',
  properties.bootstrap.server = 'ip1:port1,ip2:port2',
  privatelink.name = 'connection_name',
  privatelink.targets = '[{"az": "az1", "port": 9001}, {"az": "az2", "port": 9002}]',
  ...
);
```
For example, the traffic from broker` ip1:port1` will route to the listening port 9001 on endpoint service in AZ1.
cc @tabVersion @fuyufjh @mikechesterwang

fuyufjh (approved):

PR #46, opened 2023-02-02
RFC: connector node release workflow

tabVersion:
preview: https://github.com/risingwavelabs/rfcs/blob/tab/release-sink/rfcs/0046-release-connector-node.md

### DIFF
--- a/rfcs/0046-release-connector-node.md
+++ b/rfcs/0046-release-connector-node.md
@@ -1,0 +1,39 @@
+---
+feature: release flow for connector node
+authors:
+  - "Bohan Zhang"
+start_date: "2023/02/02"
+---
+
+# Release Workflow for Connector Node
+
+## Summary
+
+We will release the connector node as a standalone Docker image, as well as part of the all-in-one image.
+
+## Motivation
+
+The connector node is in a separate repo, and it's a standalone service but the service should run in any deployment scenario, including playground.
+
+## Design
+
+I propose to integrate the connector node's release flow into Risingwave's.
+
+### Release Workflow
+
+We publish connector node images, including nightly and stable, together with Risingwave's.
+
+* nightly builds: clone the latest code from the master branch, build and publish the image.
+* stable builds
+  * tag the code in connector node repo with a version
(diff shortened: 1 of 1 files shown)


rfcs/0046-release-connector-node.md
```diff
@@ -0,0 +1,39 @@
...
+  * in the pipeline, checkout connector node's code to the tagged version, build and publish the image
+* bug fix in connector node: need to manually trigger release workflow in the main repo
+
+### Boot-up
+
+Connector node need to start together with Risingwave, including playground, docker compose and k8s.
+
+* playground: connector node runs as a process in the same container as Risingwave.
+* docker compose: connector node will run as a separate container. It will behave like a standalone node.
+* k8s: connector node will run as a sidecar deployed in the same pod as other components, such as compute nodes and meta nodes.
```
arkbriar, line 39: Though this has been discussed a lot, I still insist on my thought: it would be better to deploy the connector nodes as standalone Pods. And it won't cost much to achieve this.

fuyufjh: In my initial design, I tent to consider the connector-node simply as **a Java-extension of the Rust process**. The only reason we introduce it is to connector RisingWave to more kinds of data systems, which is a forte of Java but hard for Rust's ecosystem.

Thus, I hope to keep the that kind of purity and never make it a *cluster* or *service*. Once you do break this, things could become much more complex - other engineers will begin to ask you: as a cluster why it can't be stateful? Why is cannot have its own monitoring and dashboard? Or even operator? It might be hard to stop them doing such things, because they sounds so reasonable!

That's why I insist on the sidecar deployment. If necessary, we should use IPC (e.g. Unix socket or Unix pipe) instead of RPC to emphasize this principle.

xxchan (commented):
Since connector code is merge into RW repo, this RFC seems outdated. Is there anything special about connector node's release workflow now?

fuyufjh:
> Since connector code is merge into RW repo, this RFC seems outdated. Is there anything special about connector node's release workflow now?

+1. We are using one repo and all-in-one image, so the process become quite straight-forward now.

PR #48, opened 2023-02-07
RFC: on conflict clause

st1page:


rfcs/0048-on-conflict-clause-in-create-table-statement.md
```diff
@@ -0,0 +1,121 @@
...
+  Notice: the specific grammar for this has not designed(e.g. should we expose append-only table to user?) but the situation is the same.
+
+- **Partial Update**
+  The ON CONFLICT make users can do partial update on the table.
+
+  ```SQL
+    CREATE table t(k int primary key, int a, int b) 
+    ON CONFLICT DO UPDATE SET (a,b) = (
+      CASE when EXCLUDED.a is not null then EXCLUDED.a else a end,
+      CASE when EXCLUDED.b is not null then EXCLUDED.b else b end);
```
xiangjinwu, line 34: nit:
```suggestion
      COALESCE(EXCLUDED.a, a),
      COALESCE(EXCLUDED.b, b));
```

rfcs/0048-on-conflict-clause-in-create-table-statement.md
```diff
@@ -0,0 +1,121 @@
...
+    Insert INTO t VALUES (1, 3);
+
+    dev=# select * from t;
+     k | cnt 
+    ---+-----
+     1 |   5
+    (1 row)
+  ```
+
+  The most strange part is that the input parameter and return value of the "reduce function" must have the same schema. But it can be easily workaround with constant expressions.
```
wcy-fdu, line 57: Just curious about how schema change will handle this case.

st1page: I think the mechanism to handle the outdated schema has no conflict with this RFC. :thinking: We have the same issue before we implement this RFC

rfcs/0048-on-conflict-clause-in-create-table-statement.md
```diff
@@ -0,0 +1,121 @@
...
+  The ON CONFLICT make users can do partial update on the table.
+
+  ```SQL
+    CREATE table t(k int primary key, int a, int b) 
+    ON CONFLICT DO UPDATE SET (a,b) = (
+      COALESCE(EXCLUDED.a, a),
+      COALESCE(EXCLUDED.b, b));
+  ```
+  
+  This feature will be useful when user use wide-table schema data model.
```
fuyufjh, line 37: As I understand it, this is similar to Apache Doris' pre-aggregate data model, where the data of the same PK can be automatically merged according to some custom rules. However, as one of its selling points, Doris has a specially designed storage to do this pre-aggregation efficiently, but we don't have it.

Apparently, our storage is optimized for non-conflict use cases, which means that most rows will not have a conflict on PK, and the PK check in MaterializeExecutor is just a precaution to avoid a crash. On the contrary, this RFC assumes that incoming rows will have many conflicts by design, and hopes to resolve them in MaterializeExecutor. I don't think it's a good idea to encourage users to do this, and it's not worth optimizing for.

Also, I am not very confident that checking in MaterializeExecutor is a good idea. Maybe it will become a painpoint at some point and maybe users will ask us to remove it. I don't really recommend relying on it much.

st1page: Anyway, we can just treat the check in the MaterializeExecutor as a pluggable feature. Now it is implemented as a flag in MaterializeExecutor, so if users ask us to remove it we can remove it easily.

rfcs/0048-on-conflict-clause-in-create-table-statement.md
```diff
@@ -0,0 +1,121 @@
...
+authors:
+  - "st1page"
+start_date: "2023/2/1"
+---
+
+# ON CONFLICT Clause in Create Table Statement
+
+## Summary
+
+allow user to declare the conflict behavior when the newly inserted row break the unique constraint of primary key.
```
twocode, line 12: by the way, when do we do constraint checks? every time a row is materialized?

st1page: Yes, the logic will be implemented in the materialize executor of the table.

rfcs/0048-on-conflict-clause-in-create-table-statement.md
```diff
@@ -0,0 +1,121 @@
...
+                  } [, ...]
+              [ WHERE condition ]
+```
+
+- **Defined when `CREATE TABLE` instead of `INSERT`**
+  Why:
+  1. Different with PG, the conflicts are not only happens in `INSERT` statement but also on the data from external source.
+  2. In PG, users can define different behavior on different `INSERT` statement. We can not implement it easily because RW treat DML statement with just one `Materialize` operator as a streaming job.
+  
+  By the way, for the `UPDATE` statement, we can just override the old value and it can not break the PK constrain after we ban the update on PK columns <https://github.com/risingwavelabs/risingwave/issues/7710>.
```
BugenZhao, line 103: PK constrain -> PK constraint

rfcs/0048-on-conflict-clause-in-create-table-statement.md
```diff
@@ -0,0 +1,121 @@
...
+
+- **DO OVERWRITE behavior**
+  just a Syntactic sugar because it is a usual usage. It has the same meaning with `DO UPDATE SET (a, b, c, ...) = (excluded.a, excluded.b, excluded.c, ...)`
+
+- **Sub-query/sub-SELECT in UPDATE**
+  It is too complex and we need implement stream join plan for the `CREATE TABLE`. So I think we need to put it on hold.
+
+### Future possibilities
+
+  We can support the `ON CONFLICT` clause in the Insert statement if we support attach the on conflict description for the chunk in `BatchInsert` executor and `StreamDML` executor to let `Materialize` executor know how to handle the conflict. But the on conflict clause is always needed in create table statement for the table with connector.
```
BugenZhao, line 116: There's still "stream path" between the `DML` and `Materialize`, also `Exchange` if there's user-specified PK, so eventually it seems we need to attach this "description" to each row, which sounds an overkill. 🤔

rfcs/0048-on-conflict-clause-in-create-table-statement.md
liurenjie1024, line 12:
```suggestion
Allow user to declare the conflict behavior when the newly inserted row break the unique constraint of primary key.
```

rfcs/0048-on-conflict-clause-in-create-table-statement.md
```diff
@@ -0,0 +1,121 @@
...
+  
+  This feature will be useful when user use wide-table schema data model.
+
+- **Express Reduce Function or UDAF?**
+  In fact, ON CONFLICT Clause can make user be able to do more things. User can even construct a streaming Map-reduce job with that where the primary key define the map fields and ON_CONFLICT clause define the reduce logic.
+  For example.
+
+  ```SQL
+    CREATE table t(k int primary key, int cnt) 
+    ON CONFLICT DO UPDATE SET cnt = t.cnt + EXCLUDED.cnt
```
liurenjie1024, line 45: Why we don't use aggregation call here:
```
select k, sum(cnt) from t;
```

st1page: It is just a trivial example here to show that this grammar can let user declare their own aggregation logic with SQL. And the loigic could be more complex.

rfcs/0048-on-conflict-clause-in-create-table-statement.md
st1page: > As I understand it, this is similar to Apache Doris' pre-aggregate data model, where the data of the same PK can be automatically merged according to some custom rules. However, as one of its selling points, Doris has a specially designed storage to do this pre-aggregation efficiently, but we don't have it.

https://doris.apache.org/docs/data-table/data-model/#aggregate-model just supports some limited behaviors. 
```
AggregationTypecurrently has the following four ways of aggregation:
SUM: Sum, multi-line Value accumulation.
REPLACE: Instead, Values in the next batch of data will replace Values in rows previously imported.
MAX: Keep the maximum.
MIN: Keep the minimum.

``` 
In fact, if we just want to support this behavior, RisingWave has already been able to express it with a `Create Materialized view` with an aggregate query. 
The `on conflicts` can take much more expressiveness to users.

rfcs/0048-on-conflict-clause-in-create-table-statement.md
```diff
@@ -0,0 +1,121 @@
...
+
+Currently, for a table with primary key, the later operations will always overwrite the previous ones when the new operation [0017-DML.md](0017-DML.md)
+But users' requirement are various here.
+I will give some different behavior here. Notice that here the conflict behavior is not just happens when an insert DML statement happening but also in a table with external connector.
+
+- **Overwrite**
+  The most important needs of using the Overwrite behavior is to express "Upsert" operation. Some connector can not give the old value of the update operation and we need to overwrite the old value with the same pk in storage.
+
+- **Ignore**
+  The Ignore behavior make user can define a "append-only table with primary key" which maintain the primary constraint and also provide a append-only changes to downstream with good performance. This is good to express a table with an At-least-once Delivery connector where a same record can be replayed multiple times to break the primary key constraint.
```
fuyufjh, line 24: Are we talking about `CREATE TABLE`? As we all agreed, tables must be updatable, so "append-only changes" here do not make senses. (because we must allow user to revise data)

rfcs/0048-on-conflict-clause-in-create-table-statement.md
```diff
@@ -0,0 +1,121 @@
...
+and conflict_action is one of:
+    DO NOTHING
+    DO UPDATE SET { column_name = { expression | DEFAULT } |
+                    ( column_name [, ...] ) = [ ROW ] ( { expression | DEFAULT } [, ...] ) |
+                    ( column_name [, ...] ) = ( sub-SELECT )
+                  } [, ...]
+              [ WHERE condition ]
+```
+
+- **Defined when `CREATE TABLE` instead of `INSERT`**
```
fuyufjh, line 98: I think we can consider this as a config for the data from connector. i.e. What command the connector should use to write data - REPLACE or INSERT or anything else.

rfcs/0048-on-conflict-clause-in-create-table-statement.md
st1page:
```SQL
create table t(v int, k int) APPEND ONLY; -- we will not expose this to user

create source s(v int, k int);
create materialized view mv as select * distinct on(k) from s;
```

st1page:
details and description in the RFC is out-of-date and the behavior of the feature is written in https://github.com/risingwavelabs/risingwave-docs/pull/1860

PR #51, opened 2023-02-14
RFC: Stream Executor with Emit on Window Close Semantics

st1page:


rfcs/0051-emit-on-close-stream-executor.md
```diff
@@ -0,0 +1,143 @@
...
+In this RFC, we will focus on how to implement a stream executor with EMIT ON WINDOW CLOSE flag and reuse the logic of the normal stream executor.
+
+## Sort Buffer Design
+
+Compared with the normal stream executor, the EMIT ON CLOSE executor needs to buffer the data until the result is complete. A general component `SortBuffer` is introduced here for executors. 
+SortBuffer consists of two parts, SortBufferCache and SortBufferTable. 
+- SortBufferTable
+  - A stateTable to persist the buffered data. 
+  - The primary key's first column is a watermark column. We call it as **sort key**
+- SortBufferCache
```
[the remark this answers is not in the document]
st1page: Yes, but it can also get from the state table when cache misses happen.

rfcs/0051-emit-on-close-stream-executor.md
```diff
@@ -0,0 +1,143 @@
...
+      - delete the row in stateTable
+### SortAgg (with batch only agg calls)
+  Under "EMIT ON CLOSE" semantics, if there is watermark in group key. We can use SortExecutor for the input stream, and then we get a sorted stream which can be processed by "BatchSortAgg". Then, we can support all the aggregators which can run in batch mode.
+  The SortAgg supports more kinds of aggregators under  "EMIT ON CLOSE", but should materialized all the input data, which could be worse than the normal streamGroupAgg.
+
+### GroupAgg
+  If there is watermark in group key, the EMIT ON CLOSE GroupAgg calculate the aggregation with the same logic with streamGroupAgg, But it only emit the complete part of the agg result. A simple idea is add a sortExecutor after a normal streamGroupAgg and we can find that the streamGroupAgg's result table and sortExecutor's stateTable exactly have the same schema and meaning. 
+  So the EMIT ON CLOSE streamGroupAgg need only add a SortBufferCache on the agg's result table.
+  - when stream chunk arrive
+    - do the same with normal streamGroupAgg
```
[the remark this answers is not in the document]
st1page: No. The `SortBuffer` here is to materialize the "output" or "result" of the  `GroupAgg` instead of the input.

rfcs/0051-emit-on-close-stream-executor.md
```diff
@@ -0,0 +1,143 @@
...
+## Summary
+
+Introduce a new component Sort Buffer used in the streaming executor and give a general method to implement EMIT ON CLOSE stream executor. Sort Buffer can support materialize and persistent the changes stream and drain the "stable" records with the input watermark. Because the input watermarks are always monotonically increasing, the output records are ordered by the watermark column and append-only.
+
+***Note: In this RFC, we will focus on the EMIT ON CLOSE query***
+
+## Background and Definitions 
+Let's look back on the current designs and RFCs. 
+
+The [RFC: The Semantics of EMIT ON WINDOW CLOSE](https://github.com/risingwavelabs/rfcs/pull/30) has given a strict definition of the semantics of **EMIT ON WINDOW CLOSE**. In short, the streaming job with the "EMIT ON CLOSE" modifier emits the result when the result is complete, and no sooner, no later. In other words, the "EMIT ON CLOSE" streaming query gives an append-only result as soon as possible.
```
TennyZhuang, line 19: `EMIT ON WINDOW CLOSE` and `EMIT ON COMPLETE` are different things, the former will ensure that a window is closed, and the latter only ensure that the result is append-only.

rfcs/0051-emit-on-close-stream-executor.md
```diff
@@ -0,0 +1,136 @@
...
+- when watermark arrive
+  - pass watermark into SortBufferCache
+- when barrier arrive
+  - drain complete rows from SortBufferCache, for each rows
+    - emit the row downstream
+    - delete the row in stateTable
+
+### SortAgg (with batch only agg calls)
+
+  Under "EMIT ON CLOSE" semantics, if there is watermark in group key. We can use SortExecutor for the input stream, and then we get a sorted stream which can be processed by "BatchSortAgg". Then, we can support all the aggregators which can run in batch mode.
```
fuyufjh, line 102:
```suggestion
  Under "EMIT ON CLOSE" semantics, if there is watermark in group key, we can use SortExecutor for the input stream, and then we get a sorted stream which can be processed by "BatchSortAgg". Then, we can support all the aggregators which can run in batch mode.
```

rfcs/0051-emit-on-close-stream-executor.md
```diff
@@ -0,0 +1,136 @@
...
+- when watermark arrive
+  - pass watermark into SortBufferCache
+- when barrier arrive
+  - drain complete rows from SortBufferCache, for each rows
+    - emit the row downstream
+    - delete the row in stateTable
+
+### SortAgg (with batch only agg calls)
+
+  Under "EMIT ON CLOSE" semantics, if there is watermark in group key, we can use SortExecutor for the input stream, and then we get a sorted stream which can be processed by "BatchSortAgg". Then, we can support all the aggregators which can run in batch mode.
```
fuyufjh, line 102: Another missing piece is the state. As we know the batch aggregate functions don't have support for states, as a result, IIUC, in order to reuse them in `SortAgg` you have to keep all these input rows in `SortAgg`'s state table and call the batch aggregate function at once to emit result. This may not be a good trade-off, especially for the value-state aggregate functions e.g. `count()`, `sum()`.

fuyufjh (commented):
The latter half of this document does not conform to Markdown format. (It requires 2 line breaks to open a new paragraph)

rfcs/0051-emit-on-close-stream-executor.md
st1page: No, actually, we do not need the aggregate functions' state here. Each group's data are totally materialized in the sort buffer table and when they are consumed in some epoch, it is a limited append only stream

st1page: Yes, so we have another EMIT ON CLOSE GroupAgg implementation as described below. I think the SortAgg is just to solve some aggregation functions which does not have a streaming version.

fuyufjh: > I think the SortAgg is just to solve some aggregation functions which does not have a streaming version(hard to implement or with high cost)

I see. Makes sense.

rfcs/0051-emit-on-close-stream-executor.md
```diff
@@ -0,0 +1,158 @@
...
+
+### SortAgg (with batch only agg calls)
+
+Under "EMIT ON WINDOW CLOSE" semantics, if there is watermark in group key, we can use SortExecutor for the input stream, and then we get a sorted stream which can be processed by "BatchSortAgg". Then, we can support all the aggregators which can run in batch mode.
+
+The SortAgg supports more kinds of aggregators under "EMIT ON WINDOW CLOSE", but should materialized all the input data, which could be worse than the normal streamGroupAgg. So we have another EMIT ON WINDOW CLOSE GroupAgg implementation as described below. I think the SortAgg is just to solve some aggregation functions which does not have a streaming version(hard to implement or with high cost)
+
+### GroupAgg
+
+If there is watermark in group key, the EMIT ON WINDOW CLOSE GroupAgg calculate the aggregation with the same logic with streamGroupAgg, But it only emit the complete part of the agg result. A simple idea is add a sortExecutor after a normal streamGroupAgg and we can find that the streamGroupAgg's result table and sortExecutor's stateTable exactly have the same schema and meaning.
```
fuyufjh, line 108: It seems to be unacceptable to me to use 2 caches in memory, which effectively means the memory is cut into 1/2, as well as the 2x IO cost of cache miss.

fuyufjh: I think the SortBufferCache here can be removed. It doesn't worth to maintain a full row copy in memory just for emitting when a `Window` event is received.

rfcs/0051-emit-on-close-stream-executor.md
```diff
@@ -0,0 +1,158 @@
...
+
+[RFC: Over Window on watermark](https://github.com/risingwavelabs/rfcs/pull/8)
+We still need more detailed design here, but it is clear that the row's complete and row's emit can not always happens at the same time.
+
+### EqJoin
+
+It likes window join in Flink.
+When there are watermarks in the first join key of the both two sides. We can use two SortBuffers(table and cache) to implement a sortMergeJoin. SortBuffer’s `peek` method is used here.
+
+### IntervalJoin
```
fuyufjh, line 136: Same thing here. Let me read the docs of Flink again to see whether there are better ideas...

fuyufjh (commented):
.

chenzl25:
With emit on close semantics, it seems we can implement event time temporal join as well, because this semantics can ensure all the data we need are complete.

rfcs/0051-emit-on-close-stream-executor.md
fuyufjh: See https://github.com/risingwavelabs/rfcs/pull/32#discussion_r1059266190

In short, Flink puts join key before time column, rather than 2 copies of data. I think we may start with it first so that hopefully it can reuse the source code of the normal version of interval join operator.

rfcs/0051-emit-on-close-stream-executor.md
st1page: Yes, what you say makes sense here, because every record will just be read one time and then will be deleted soon. The cache is meaningless here.

fuyufjh (approved):

PR #52, opened 2023-02-15
RFC: Create Sink into Table

st1page:


rfcs/0052-create-sink-into-table.md
```diff
@@ -0,0 +1,41 @@
...
+---
+
+# Create Sink into Table
+
+## Motivation
+
+### Loss of functionality compared to FlinkSQL
+FlinkSQL uses the `INSERT ... SELECT ...` statement to create a streaming job. This means that in FlinkSQL, the table content is not bound to the result of a streaming query, even if the table is stored in Flink table store (https://nightlies.apache.org/flink/flink-table-store-docs-master/). **Users can sink more than one stream into a table and let the table handle the pk conflict by itself**.
+
+However, RisingWave uses `create materialised view as query' to declare a streaming query and bind a table with the query's result set. So compared to FlinkSQL, RisingWave cannot express multiple streaming query into one internal table naturally.
```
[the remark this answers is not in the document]
liurenjie1024: Honestly speaking, I don't think this is a good practice, e.g. multi insert statement into one table. Why not just use union/join in one select query? Also, this may introduce many problems such as concurrent modification error in lake house format.

rfcs/0052-create-sink-into-table.md
```diff
@@ -0,0 +1,41 @@
...
+
+### RisingWave's own expressive enhancement
+It can also improve RisingWave's own expressiveness even if we actually do not want to sink multiple stream into one table. using the ON CONFLICT clause ([RFC: on conflict clause](https://github.com/risingwavelabs/rfcs/pull/48)) and `Create sink into table` statement, user can define some user-defined aggregation logic if we have not implemented [User-Defined Aggregates](https://www.postgresql.org/docs/current/xaggr.html).
+The user-defined aggregates feature is also needed by some POC users. If we implement `on conflict clause' and `create sink into table', we can temporarily put UDAF on hold. I think UDAF is a bit hard in streaming because users should define their own state storage in the Aggregates function so that RisingWave can persist them.
+
+## Design
+### Syntax and semantics
+```SQL
+  CREATE SINK INTO [table_name] AS select_query
+```
```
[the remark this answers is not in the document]
st1page: Yes and I think the "MV" here is the table defined by user in that case.

rfcs/0052-create-sink-into-table.md
```diff
@@ -0,0 +1,399 @@
...
+
+Here are some Q&A for the details of the graph
+- What's the difference between "append only table" and "table without pk"?
+  - The only difference is that if Update/delte on the table is allowed. which make the non-append only table without pk
+    - not append only, so the watermark is not allowed on it.
+    - the DML changes must be hash shuffled to the materialize fragment.
+    - the materialize must handle pk conflict for the DML changes.
+  - maybe we will support alter table between append only and non-append only in future.  
+- Why do we need blackhole SinkExecutor here? 
+  - we have some special logic in the sink executor such as `force-append-only` and special compaction when the down stream's pk does not contain stream key, we can reuse these logic in SinkExecutor to maintain the same behvaior with other kind of sinks.
```
fuyufjh, line 378: I think this Executor is far different from a `SinkExecutor`. `SinkExecutor` has no output channel, and focused on executor the Sink of a particular external connector.

Actually, I think `force-append-only` is too simple and is not worth of reusing. And if you do want to reuse it, perhaps we can try to extract these logics into a new executor. Either way, I would definitely -1 for reusing `SinkExecutor`.

rfcs/0052-create-sink-into-table.md
```diff
@@ -0,0 +1,399 @@
...
+The statement requires **Allow a sink and a table have the same name**. It is ok in our system because the sink will never appear in the FROM clause. An alternative is that we can add a `_sink` suffix on the sink's name.
+
+### Implementation
+
+Thanks for @BugenZhao and @shanicky's offline discussion.
+#### CREATE SINK into TABLE
+
+Currently, the table's plan is shown as the graph. In brief, the table's data can comes from the external source or BatchDML. And the DMLExec is used here as a side input. ![](./images/0052-sink-to-table/cur-table-plan.png).
+
+The `create sink to table` introduces more data sources for the table. And now RW has implemented stream Union executor and RowIdGen with specified distribution. So we can support it more clearly with these new features.
```
fuyufjh, line 367: > stream Union executor

IIUC, you mean merging the events from multiple upstream into one stream? If so, please call it MergeExecutor or whatever. It's much different from `Union` in relational algebra.

rfcs/0052-create-sink-into-table.md
```diff
@@ -0,0 +1,399 @@
...
+    - the DML changes must be hash shuffled to the materialize fragment.
+    - the materialize must handle pk conflict for the DML changes.
+  - maybe we will support alter table between append only and non-append only in future.  
+- Why do we need blackhole SinkExecutor here? 
+  - we have some special logic in the sink executor such as `force-append-only` and special compaction when the down stream's pk does not contain stream key, we can reuse these logic in SinkExecutor to maintain the same behvaior with other kind of sinks.
+- "Exchange Any"
+  - It appears on the append only stream whose row_id column has not been generated. So any uniform shuffle strategy is ok here. 
+- Why not move the ProjectExec for the generated column on the Materialize's fragment?
+  - when there is a user defined pk on the table. The pk might contains generated column, which will be used to shuffle. So it must be generated before the exchange executor.
+  - when there is no pk, we can move it on the MaterializeExecutor's fragment. But I prefer make it consistent with the case having pk.
```
fuyufjh, line 383: How about moving it into `DMLExecutor`?

rfcs/0052-create-sink-into-table.md
```diff
@@ -161,7 +150,82 @@ GROUP BY window_start;
...
+    price BIGINT,
+    event_time TIMESTAMP,
+) APPEND ONLY;
+
+CREATE TABLE dws_product (
+  product VARCHAR,
+  order_cnt BIGINT,
+  PRIMARY KEY (product)
+) WITH (
+  'ttl' = '7d'
```
fuyufjh, line 184: `ttl` is still dangerous and should not be used by common users 🥲

rfcs/0052-create-sink-into-table.md
```diff
@@ -161,7 +150,82 @@ GROUP BY window_start;
...
+  -- only update when the new incoming row's order_cnt is larger than the old order_cnt
+) ON CONFLICT DO UPDATE SET (product, order_cnt) = (
+      CASE WHEN EXCLUDED.order_cnt IS NOT NULL AND EXCLUDED.order_cnt > order_cnt 
+        THEN EXCLUDED.product ELSE product 
+      END,
+      CASE WHEN EXCLUDED.order_cnt IS NOT NULL AND EXCLUDED.order_cnt > order_cnt 
+        THEN EXCLUDED.order_cnt, order_cnt ELSE order_cnt 
+      END)
```
fuyufjh, line 192: `ON CONFLICT DO UPDATE SET column1 = <expression1>, column2 = <expression2>, ...` looks better to me. 

Reference: https://www.postgresql.org/docs/current/sql-insert.html

rfcs/0052-create-sink-into-table.md
```diff
@@ -161,7 +150,82 @@ GROUP BY window_start;
...
+  product,
+  COUNT(*) as order_cnt,
+FROM dwd_orders
+GROUP BY product
+WITH (
+  -- because the table's on- conflict behavior can only applied on the “INSERT” conflict，so we need use force append-only sink here.
+  force_append_only = 'true'
+);
+
+-- when the user want to add an new aggregator
```
fuyufjh, line 206: I don't really like this use case. It basically adds another identical streaming job (near 2x cost) just to put a new aggregate function there. @BugenZhao's design (TODO: not published to RFC repo yet) would be a better approach.

rfcs/0052-create-sink-into-table.md
```diff
@@ -0,0 +1,399 @@
...
+
+-- when the new job catch up
+DROP SINK order_production_analyze_per_hour_job;
+ALTER SINK new_order_production_analyze_per_hour_job RENAME TO order_production_analyze_per_hour_job;
+```
+In fact the complex ON CONFLICT behavior and the `sql_version` column are just because we can not add a filter condition on the exist streaming job, It can be replaced by "alter mv" feature too.
+
+```SQL
+ALTER TABLE dws_product ADD sell_sum BIGINT;
+ALTER order_production_analyze_per_hour_job AS SELECT
```
fuyufjh, line 299:
```suggestion
ALTER SINK order_production_analyze_per_hour_job AS SELECT 
```

rfcs/0052-create-sink-into-table.md
st1page: `( column_name [, ...] ) = [ ROW ] ( { expression | DEFAULT } [, ...] ) |` is supported by PG too. Btw,the too many same CASE WHEN condition are really annoying, which make me a little want to do some hack like
```
(column_name [, ...] ) CASE 
 WHEN conditon1 THEN (expression11, expression12, ...)
 WHEN conditon2 THEN (expression21, expression22, ...)
END
```

rfcs/0052-create-sink-into-table.md
st1page: The generated column Project is needed for the both three path of the table: DML, external source and Sink to the table. we can not just handle it in the `DMLExecutor`

rfcs/0052-create-sink-into-table.md
st1page: Agree, this table's ttl will introduce wrong check pk behavior and give the downstream an illegal changes stream

st1page: Agree, this table's TTL will make the materialize executor wrongly check pk and give the downstream an illegal changes stream

rfcs/0052-create-sink-into-table.md
st1page: Yes, but as a streaming executor they are same 🤔

rfcs/0052-create-sink-into-table.md
st1page: In fact, the sink executor has output and can give downstream a processed stream. I guess we will have more and more common logic in the sink executor such as https://risingwave-labs.slack.com/archives/C048771BGKF/p1693553432021789.
 So. I think a blackhole sink here makes sense.

rfcs/0052-create-sink-into-table.md
fuyufjh: Oh, that's true. `DMLExecutor` is not a side-input anymore in this design.

Here the major problem of existing implementation (https://github.com/risingwavelabs/risingwave/pull/11698) is that we all know it's not a usual `ProjectExecutor` because it should not generate values for update & delete events. It doesn't matter whether to reuse an existing executor (https://github.com/risingwavelabs/risingwave/issues/11738) or create a new one.

So I would suggest placing a new operator here, at least for the optimizer stage. On executor stage, we simply accomplish this by adding a few lines in `ProjectExecutor`.

rfcs/0052-create-sink-into-table.md
```diff
@@ -0,0 +1,400 @@
...
+WITH (
+  force_append_only = 'true'
+);
+
+-- when the new job catch up
+DROP SINK order_production_analyze_job;
+ALTER SINK new_order_production_analyze_job RENAME TO order_production_analyze_job;
+```
+
+2. Apply delta
```
BugenZhao, line 229: This also looks too complicated to me. All these 100 lines of SQL achieve the exactly same stuff as `CREATE OR REPLACE MATERIALIZED VIEW AS <add new agg call>`, where I don't think there's any need to expose such details to users.

BugenZhao: On the other side, the new sink doesn't have access to the states in the old sink, so this can be hardly applied to stateful queries.

rfcs/0052-create-sink-into-table.md
fuyufjh: Not exactly. It's more like a continuous version of Chain operator. Optimization stuff like PK derivation, selectivity estimation, etc. also differ, although the optimizer won't process it for now.

My proposal is that we can introduce a `Merge` operator in Optimizer side only, and both `Merge` and `Union All` will be converted to the existing `UnionExecutor` (which might be renamed to avoid ambiguity with `Union` in SQL)

rfcs/0052-create-sink-into-table.md
fuyufjh: Hmm, I remember that the reason why "the sink executor has output" is because that makes sink executors **non-special** - it's driven by the actor's `for` loop just like any other executors.

It's like using an undefined behavior, that's why I feel not so good.

BugenZhao (commented):
From my very own perspective, I'm neutral about the introduction of this syntax. While it brings better flexibility in managing streaming pipelines in our system and makes the big-data guys happy to interact with our system, it also undermines the unity of the ecosystem that we establish through materialized views: users will always face the dilemma of choosing a pattern when building a new streaming job.

However, I'm strong -1 for the SQL evolution approach based on sinking into the table, at least as the best practice we recommend users to do. This part is really not the style that how RisingWave handles the streaming problems.

rfcs/0052-create-sink-into-table.md
```diff
@@ -0,0 +1,400 @@
...
+  product,
+  NULL as click_cnt,
+  COUNT(*)
+FROM TUMBLE(dwd_orders, order_time, INTERVAL '1' HOUR)
+GROUP BY window_start, product;
+```
+
+### provide user more control over the result table
+- Do DML on it and do modifications
+- define Primary key and on conflict behavior
```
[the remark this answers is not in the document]
st1page: I think it is an independent issue with creating sink to the table. For a sink or sink executor, if the down stream pk and its stream key is different, It should solve this issue anyway. It does not depend on the target table is RW's table or a external table.

rfcs/0052-create-sink-into-table.md
```diff
@@ -0,0 +1,400 @@
...
+
+### alter mv to table
+
+To make it more usable for users, I would like to introduce another function to alter an exiting materialzied view to a table.
+
+```sql
+ALTER MATERIALIZED VIEW mv_table TO TABLE
+```
+
+This is because now we still recommend `create materialized view` at the first place instead of `create table as`. However, application developers are facing changes all the time, so it's very likely that they may realize they need to 'alter' an existing materialized view. Meanwhile, the strong consistency gurantee of MV prevents us to do any altering. With the help of `ALTER MATERIALIZED VIEW TO TABLE`, users can choose to break this consistency gurantee and alter their streaming pipelines.
```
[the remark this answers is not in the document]
st1page: LGTM.

rfcs/0052-create-sink-into-table.md
```diff
@@ -0,0 +1,400 @@
...
+```
+
+### SQL evolution
+After decoupling the sink and the table. Users have chance to change the sink's SQL by drop the old sink and rebuild the new sink to the table. In that case, the table's downstream will not be affected. 
+But RW can not give consistent guarantee, so user must give additional declaration in SQL.
+
+I summarize two patterns here
+- Backfill: If user have all historical data in the upstream table, he can drop the old the create the new sink to the table
+  - 👍 Easy to understand. No more SQL rewritten
+  - 👍 eventually consistency
```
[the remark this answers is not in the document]
st1page: Just in the next 3rd line...

rfcs/0052-create-sink-into-table.md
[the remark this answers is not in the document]
BugenZhao: Yeah, you're right. Just as a reminder.

rfcs/0052-create-sink-into-table.md
st1page: > On the other side, the new sink doesn't have access to the states in the old sink, so this can be hardly applied to stateful queries.

I guess this restriction is requiring a event time column in the result table so that user can divide their result set :thinking:

rfcs/0052-create-sink-into-table.md
st1page: Emmm, so how about requiring the sink executor's output should be the data that the sink should emit? What logic I want to reuse is "parse the sink catalog and do some process on the input stream". I guess this should be done in the sink executor.

rfcs/0052-create-sink-into-table.md
```diff
@@ -0,0 +1,400 @@
...
+ALTER MATERIALIZED VIEW mv_table TO TABLE
+```
+
+This is because now we still recommend `create materialized view` at the first place instead of `create table as`. However, application developers are facing changes all the time, so it's very likely that they may realize they need to 'alter' an existing materialized view. Meanwhile, the strong consistency gurantee of MV prevents us to do any altering. With the help of `ALTER MATERIALIZED VIEW TO TABLE`, users can choose to break this consistency gurantee and alter their streaming pipelines.
+
+Currently, we can just support it with limitation that the mv's primary key can not be hidden column, we will discuss why in the "implementation" phrase.
+
+The statement requires **Allow a sink and a table have the same name**. It is ok in our system because the sink will never appear in the FROM clause. An alternative is that we can add a `_sink` suffix on the sink's name.
+
+### Implementation
```
st1page, line 361:
```suggestion
## Implementation
```

rfcs/0052-create-sink-into-table.md
fuyufjh: > "parse the sink catalog and do some process on the input stream"

👍 It makes more sense to me.

rfcs/0052-create-sink-into-table.md
```diff
@@ -0,0 +1,400 @@
...
+This RFC give a new SQL grammar `CREATE SINK sink_name TO table_name AS SELECT ...` to create streaming job and apply its changes into a RW's table, which decouples the streaming job and the result table.
+- It's changes events behavior is same with other external sinks.
+- When table's constraint is violated by the changes, the behavior should be determined by the RW's table. Specifically, it can be defined by ON CONFLICT clause ([RFC: on conflict clause](https://github.com/risingwavelabs/rfcs/pull/48)). 
+- align the barrier and epoch with upstream, satisfying exactly-once and snapshot isolation with upstream.
+
+Also, the `ALTER MATERIALIZED VIEW [mv_table] TO TABLE` is used for transforming an existent MV to a table and a Sink to the table. 
+## Motivation
+There are some use cases.
+
+### Multiple data streams into a single table
```
fuyufjh, line 36: Currently, AFAIK, more users want to merge multiple data streams row-by-row (similar to an `union all`) instead of column-by-column.

rfcs/0052-create-sink-into-table.md
```diff
@@ -0,0 +1,400 @@
...
+
+```sql
+ALTER MATERIALIZED VIEW mv_table TO TABLE
+```
+
+This is because now we still recommend `create materialized view` at the first place instead of `create table as`. However, application developers are facing changes all the time, so it's very likely that they may realize they need to 'alter' an existing materialized view. Meanwhile, the strong consistency gurantee of MV prevents us to do any altering. With the help of `ALTER MATERIALIZED VIEW TO TABLE`, users can choose to break this consistency gurantee and alter their streaming pipelines.
+
+Currently, we can just support it with limitation that the mv's primary key can not be hidden column, we will discuss why in the "implementation" phrase.
+
+The statement requires **Allow a sink and a table have the same name**. It is ok in our system because the sink will never appear in the FROM clause. An alternative is that we can add a `_sink` suffix on the sink's name.
```
fuyufjh, line 359: > An alternative is that we can add a `_sink` suffix on the sink's name.

+1. Keep it simple and stupid

fuyufjh:
The RFC actually contains multiple tasks. Personally I think the priority might be

1. **create sink into table**
2. **create table as query** (because it's a syntax sugar based on 1)
3. **alter MV to table** (because it can be workarounded by create sink on that MV)
4. **ON CONFLICT PARTIAL UPDATE** (depending on user requirements)
5. **on conflict do update set** (depending on user requirements that 4 can't fulfill)

rfcs/0052-create-sink-into-table.md
```diff
@@ -0,0 +1,400 @@
...
+  - maybe we will support alter table between append only and non-append only in future.  
+- Why do we need blackhole SinkExecutor here? 
+  - we have some special logic in the sink executor such as `force-append-only` and special compaction when the down stream's pk does not contain stream key, we can reuse these logic in SinkExecutor to maintain the same behvaior with other kind of sinks.
+- "Exchange Any"
+  - It appears on the append only stream whose row_id column has not been generated. So any uniform shuffle strategy is ok here. 
+- Why not move the ProjectExec for the generated column on the Materialize's fragment?
+  - when there is a user defined pk on the table. The pk might contains generated column, which will be used to shuffle. So it must be generated before the exchange executor.
+  - when there is no pk, we can move it on the MaterializeExecutor's fragment. But I prefer make it consistent with the case having pk.
+- Default column Project
+  - For the Insert statement, the default column is handled in the batchInsert executor. To achieve the same behavior for the create sink into table, we must handle it between the sink and the materialize executor.
```
fuyufjh, line 386: Shall we change the implementation of `BatchInsert` to make them consistent?

By the way, what's the difference between `default` and `generated` column? 🤔

chenzl25:
Currently, our optimizer will analyze some properties e.g. `cardinality` for Mviews, and persist them in somewhere. It is possible that other Mviews rely on this property. However, if we alter a Mview to a table, it could break this property because users can update their table in any way. Our features like dynamic filter and scalar subquery assumption would be broken.

rfcs/0052-create-sink-into-table.md
st1page: @st1page

chenzl25:
`Sink into table` requires the backfilling to have resuming ability at any time because `sink into table` would return success to users immediately and there is no way to cancel the sink in the middle. We have to finish backfilling. If a system recovery happens during backfilling in `sink to table`, we need the resuming ability to provide exactly-once semantics.

fuyufjh:
> Currently, our optimizer will analyze some properties e.g. `cardinality` for Mviews, and persist them in somewhere. It is possible that other Mviews rely on this property. However, if we alter a Mview to a table, it could break this property because users can update their table in any way. Our features like dynamic filter and scalar subquery assumption would be broken.

The only solution I can think of here is marking the depended MV as being unable to be altered to table. For example, store the list of depending MV (that prevents it from being altered to table) in its catalog, if the list is not empty, then `alter table` will be rejected.

BugenZhao:
> > Currently, our optimizer will analyze some properties e.g. `cardinality` for Mviews, and persist them in somewhere. It is possible that other Mviews rely on this property. However, if we alter a Mview to a table, it could break this property because users can update their table in any way. Our features like dynamic filter and scalar subquery assumption would be broken.
> 
> 
> 
> The only solution I can think of here is marking the depended MV as being unable to be altered to table. For example, store the list of depending MV (that prevents it from being altered to table) in its catalog, if the list is not empty, then `alter table` will be rejected.
> 
> 
> 
> Btw, is `cardinality` the only case here?

Another possible approach I can come up with might be  to trigger a series of `ALTER` recursively for downstream jobs. If they're still able to plan and the plan is compatible without those good properties, then we can claim that downstream jobs don't rely on the properties and we can safely alter it into a table.

st1page:
> Btw, is cardinality the only case here?

Watermark and append-only. may be we will add the unique key in future.

Append only will be more troublesome... we can not alter the mv with append-only max aggregation to a non-append-only max aggregation...

st1page:
> The RFC actually contains multiple tasks. Personally I think the priority might be
> 
> 1. **create sink into table**
> 2. **create table as query** (because it's a syntax sugar based on 1)
> 3. **alter MV to table** (because it can be workarounded by create sink on that MV)
> 4. **ON CONFLICT PARTIAL UPDATE** (depending on user requirements)
> 5. **on conflict do update set** (depending on user requirements that 4 can't fulfill)

Update the meeting conclusion: 
what we will do:
1. **create sink into table** (It should require and block on the backfilling recovery feature to make sure the exactly-once from the upstream mv)
2. **create table as query** (because it's a syntax sugar based on 1)
3. **ON CONFLICT PARTIAL UPDATE** (depending on user requirements)
4. **on conflict do update set** (depending on user requirements that 4 can't fulfill)

what still needs more design:
Alter mv to table has some troublesome issues
- it can not serve all situations, especially when the mv has hidden pk columns
- down stream plan could depend on the MV's logical properties analyzed by the optimizer.

rfcs/0052-create-sink-into-table.md
```diff
@@ -0,0 +1,400 @@
...
+
+A circular streaming plan is forbidden, with `dependent_relations` we can easily do it.
+
+### create table as 
+
+Additionally, we may introduce a `CREATE TABLE AS` syntax, which can be considered as a sugar of `CREATE TABLE (...)` and `CREATE SINK INTO` combined together. This syntax is also adopted by [Flink SQL](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=199541185) and [ksqlDB](https://docs.ksqldb.io/en/latest/developer-guide/ksqldb-reference/create-table-as-select/).
+
+```sql
+CREATE TABLE table_name AS select_query
+```
```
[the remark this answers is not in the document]
BugenZhao: Yes. [`SELECT INTO`](https://www.postgresql.org/docs/current/sql-selectinto.html) might be a good candidate for the syntax replacement of the original batch query.

xxchan (commented):
can merge? anything to refine?

rfcs/0052-create-sink-into-table.md
```diff
@@ -0,0 +1,400 @@
...
+    - The RW's streaming execution need additional 2pc mechanism with external storage to achieve exactly-once.
+
+This RFC give a new SQL grammar `CREATE SINK sink_name TO table_name AS SELECT ...` to create streaming job and apply its changes into a RW's table, which decouples the streaming job and the result table.
+- It's changes events behavior is same with other external sinks.
+- When table's constraint is violated by the changes, the behavior should be determined by the RW's table. Specifically, it can be defined by ON CONFLICT clause ([RFC: on conflict clause](https://github.com/risingwavelabs/rfcs/pull/48)). 
+- align the barrier and epoch with upstream, satisfying exactly-once and snapshot isolation with upstream.
+
+Also, the `ALTER MATERIALIZED VIEW [mv_table] TO TABLE` is used for transforming an existent MV to a table and a Sink to the table. 
+## Motivation
+There are some use cases.
```
xxchan, line 34: Should we add these use cases to the user doc? https://docs.risingwave.com/docs/current/sql-create-sink-into/

Currently, nothing is mentioned there. I doubt whether users can understand these abilities.

xxchan: Oh, I see https://github.com/risingwavelabs/risingwave-docs/issues/2592 mentioned guides are missing

fuyufjh (approved):

PR #53, opened 2023-02-21
RisingWave Telemetry Design

odysa:


xxchan (approved):
Implemented at https://github.com/risingwavelabs/risingwave/pull/7384

@odysa Is there anything need to be edited? If not, let's merge it.

PR #54, opened 2023-02-24
RFC: Suspend MV on Non-Recoverable Errors

hzxa21:
[Rendered Link](https://github.com/risingwavelabs/rfcs/blob/a3d8ce7f468be5a7a4f68358f4c8a3603a52d6a1/rfcs/0054-suspend-mv-on-non-recoverable-errors.md)

### DIFF
--- a/rfcs/0054-suspend-mv-on-non-recoverable-errors.md
+++ b/rfcs/0054-suspend-mv-on-non-recoverable-errors.md
@@ -1,0 +1,63 @@
+---
+feature: Suspend MV on Non-Recoverable Errors
+authors:
+  - "Patrick Huang"
+start_date: "2023/2/24"
+---
+# RFC: Suspend MV on Non-Recoverable Errors
+
+# Motivation
+
+Currently we have two error handling mechanisms on MV:
+
+1. Ignore and tolerate the compute error. Relevant issue: https://github.com/risingwavelabs/risingwave/issues/4625)
+2. Report error to meta and trigger a full recovery.
+
+However, when there are non-recoverable errors caused by malform records ingested by the user, the current error handling mechanisms are insufficient:
+
+1. Even if these messages can be skip or fixed via filling in default values, correctness can be affected, especially when sinking to external systems, since the damage may hardly get reverted.
+2. Recovery won’t help here. We may end up with re-consuming the malform records and trigger full recovery over and over again. All MVs in the cluster will be affected and no progress can be made.
+
+This RFC proposes that we should suspend the affected MV when encountering non-recovery errors. Non-recoverable errors by definition cannot be fixed automatically so manual intervention and re-creating the MV are unavoidable. The goals of this RFC are:
+
+- Reduce the radius of impact when non-recoverable errors happen. In other words, non-relevant MVs should not be affected.
+- Improve observability when non-recoverable errors happen. In other words, MV progress and necessary states should be kept to help user/SRE identify the issue.
+
+# Design
+
+## Non-Recoverable Errors
+
+- Errors caused by malform records ingested by the user source. Examples: record too large, record parsing error, record format/type mismatch.
+- Errors caused by failure in sanity check. Examples: state table insert/update/delete sanity check fails.
+
+## Suspend MV on Non-Recoverable Errors
+
+Error MV and all its downstream MVs are suspended and removed from the streaming graph for checkpointing. Suspended MV and its internal states are still queryrable (at least for SRE). You can treat the suspended MV as a dropped but not GCed MV.
+
+Detail steps on MV suspension:
+
+1. Actor reports non-recoverable errors to meta node.
+2. Meta node identifies the MV of the failed actor as well as all its downstream MVs.
+3. Meta node suspends these MVs by removing them from the streaming graph.
+4. A suspended MV will be deleted manually or automatically TTLed.
+    
+    
+
+### DISUCSSION: How to suspend the MV?
+
+1. Trigger a recovery: all ongoing concurrent checkpoint fails and actors of the suspended MV won’t get rebuilt during recovery.
+2. No recovery triggered: remove the suspended MV from the streaming graph immediately or in the next checkpoint. (Is this possible? e.g. failure in downstream MVs may cause backpressure in upstream and can we preempt backpressure?) 
+
+Personally prefer a > b.
+
+### DISCUSSION: What happen when user queries the suspended MV?
+
+1. Ban user query and throw an error.
+2. Allow users to query but remind them there is an error and the MV will not longer be updated.
+
+Personally prefer a > b. 
+
+## Future possibilities
+
+- We can store the error message, stack trace as well as other detailed information of the suspended MV in a user queryable system table.
+-


StrikeW:
I think we may extend the scenarios to unrecoverable Sources (#7192). The suspend workflow should be similar from the perspective of Meta.

rfcs/0054-suspend-mv-on-non-recoverable-errors.md
```diff
@@ -0,0 +1,63 @@
...
+
+1. Ignore and tolerate the compute error. Relevant issue: https://github.com/risingwavelabs/risingwave/issues/4625)
+2. Report error to meta and trigger a full recovery.
+
+However, when there are non-recoverable errors caused by malform records ingested by the user, the current error handling mechanisms are insufficient:
+
+1. Even if these messages can be skip or fixed via filling in default values, correctness can be affected, especially when sinking to external systems, since the damage may hardly get reverted.
+2. Recovery won’t help here. We may end up with re-consuming the malform records and trigger full recovery over and over again. All MVs in the cluster will be affected and no progress can be made.
+
+This RFC proposes that we should suspend the affected MV when encountering non-recovery errors. Non-recoverable errors by definition cannot be fixed automatically so manual intervention and re-creating the MV are unavoidable. The goals of this RFC are:
```
jon-chuang, line 21: I think this may be too strict.

The user may still want the MV to run. 

I believe that the user should be in charge of ensuring that their input source is persisted outside of RW (e.g. in msg queue like kafka). They should create a new MV that avoids the errors if they discover that they cannot tolerate the given errors.

jon-chuang: Alternately, we should indeed expose this 'strict' option to the user. However, we need to decide which mode is the default:

```yaml
connector_error_suspends_mv: true,
connector_error_kills_batch_query: true,
compute_error_suspends_mv: true,
compute_error_kills_batch_query: true,
```

There is some overlap with the batch query behaviour (whether to warn or throw error) @fuyufjh . We have not decided which is the optimal default behaviour for user.

rfcs/0054-suspend-mv-on-non-recoverable-errors.md
```diff
@@ -0,0 +1,63 @@
...
+# Motivation
+
+Currently we have two error handling mechanisms on MV:
+
+1. Ignore and tolerate the compute error. Relevant issue: https://github.com/risingwavelabs/risingwave/issues/4625)
+2. Report error to meta and trigger a full recovery.
+
+However, when there are non-recoverable errors caused by malform records ingested by the user, the current error handling mechanisms are insufficient:
+
+1. Even if these messages can be skip or fixed via filling in default values, correctness can be affected, especially when sinking to external systems, since the damage may hardly get reverted.
```
jon-chuang, line 18: > the damage may hardly get reverted.

I'm not completely sure this is correct. 

Could you provide some examples where the damage is permanent? One way iss that their downstream components to RW are affected, and a lot of manual work is required to clean up the wrong data.

For RW itself, I think we should handle by deleting the old MV and creating a new one.

Ultimately, I think the behaviour should be determined by the user, whether they are willing to tolerate certain types of errors or if they want to suspend the job. The same goes for batch queries.

hzxa21: > One way iss that their downstream components to RW are affected, and a lot of manual work is required to clean up the wrong data.

Exactly.

> Ultimately, I think the behaviour should be determined by the user, whether they are willing to tolerate certain types of errors or if they want to suspend the job. The same goes for batch queries.

Totally agree. That was my intention too.

rfcs/0054-suspend-mv-on-non-recoverable-errors.md
```diff
@@ -0,0 +1,63 @@
...
+This RFC proposes that we should suspend the affected MV when encountering non-recovery errors. Non-recoverable errors by definition cannot be fixed automatically so manual intervention and re-creating the MV are unavoidable. The goals of this RFC are:
+
+- Reduce the radius of impact when non-recoverable errors happen. In other words, non-relevant MVs should not be affected.
+- Improve observability when non-recoverable errors happen. In other words, MV progress and necessary states should be kept to help user/SRE identify the issue.
+
+# Design
+
+## Non-Recoverable Errors
+
+- Errors caused by malform records ingested by the user source. Examples: record too large, record parsing error, record format/type mismatch.
```
jon-chuang, line 30: > record too large

I know this is just an example, but for completeness, in what scenario can `record too large` occur? 

I know we have encountered request too large for etcd request: https://github.com/risingwavelabs/risingwave/issues/7728 but not sure of `record too large` issue of connectors. @tabVersion

hzxa21: We recently encounter an issue when user ingests a malform row with >64KB pk. We didn't notice it until compactor stops working and we manually inspect the encoded SSTs in S3. User also didn't realize that they have such a malform row ingested before we explicitly told them there is something wrong with a specific table. 

Ideally, we should warn the user or even stop ingesting the malform row once it appeared. Although we have a [hotfix](https://github.com/risingwavelabs/risingwave/pull/8064) merged, we still need to figure out a way to do that instead of just workaround the issue by allowing >64KB keys to be stored.

Also, we doesn't seem to have enforce any length limitation on our PK or the VARCHAR type but this is a separate discussion.

fuyufjh (commented):
Generally agree. 

Some complements to the background: Previously, the major reason blocking us from doing it is we didn't assume "Non-Recoverable Errors", because recovering a suspended streaming job seems to be difficult. But, as your solution limited the scope to "non-recoverable errors", it becomes doable.

rfcs/0054-suspend-mv-on-non-recoverable-errors.md
```diff
@@ -0,0 +1,63 @@
...
+- Errors caused by failure in sanity check. Examples: state table insert/update/delete sanity check fails.
```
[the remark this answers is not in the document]
hzxa21: > Shall we consider panic as a kind of Non-Recoverable Errors?

+1

rfcs/0054-suspend-mv-on-non-recoverable-errors.md
```diff
@@ -0,0 +1,63 @@
...
+2. No recovery triggered: remove the suspended MV from the streaming graph immediately or in the next checkpoint. (Is this possible? e.g. failure in downstream MVs may cause backpressure in upstream and can we preempt backpressure?) 
+
+Personally prefer a > b.
+
+### DISCUSSION: What happen when user queries the suspended MV?
+
+1. Ban user query and throw an error.
+2. Allow users to query but remind them there is an error and the MV will not longer be updated.
+
+Personally prefer a > b.
```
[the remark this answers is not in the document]
hzxa21: 2 is doable since the data remains on storage and it is immutable. May need some refactoring on the meta side though (maybe we can just treat the dangled MV and its internal state tables as regular batch tables)

rfcs/0054-suspend-mv-on-non-recoverable-errors.md
```diff
@@ -0,0 +1,63 @@
...
+1. Actor reports non-recoverable errors to meta node.
+2. Meta node identifies the MV of the failed actor as well as all its downstream MVs.
+3. Meta node suspends these MVs by removing them from the streaming graph.
+4. A suspended MV will be deleted manually or automatically TTLed.
+    
+    
+
+### DISUCSSION: How to suspend the MV?
+
+1. Trigger a recovery: all ongoing concurrent checkpoint fails and actors of the suspended MV won’t get rebuilt during recovery.
```
BugenZhao, line 48: IIUC, we expect the users to drop the problematic materialized views to recovery from this. However, does this mean that the materialized view is "dangled" at this time?

rfcs/0054-suspend-mv-on-non-recoverable-errors.md
```diff
@@ -0,0 +1,63 @@
...
+- Errors caused by malform records ingested by the user source. Examples: record too large, record parsing error, record format/type mismatch.
+- Errors caused by failure in sanity check. Examples: state table insert/update/delete sanity check fails.
+
+## Suspend MV on Non-Recoverable Errors
+
+Error MV and all its downstream MVs are suspended and removed from the streaming graph for checkpointing. Suspended MV and its internal states are still queryrable (at least for SRE). You can treat the suspended MV as a dropped but not GCed MV.
+
+Detail steps on MV suspension:
+
+1. Actor reports non-recoverable errors to meta node.
```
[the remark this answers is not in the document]
st1page: I think the `catch_unwind` is just to get the information about the error's place such as the actor_id and send them to meta. A check-point-based failover is still needed?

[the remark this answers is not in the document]
BugenZhao: Yes. In this case, we must rely on approach #1 of suspending the MV. 🤔

rfcs/0054-suspend-mv-on-non-recoverable-errors.md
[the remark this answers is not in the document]
fuyufjh: After some extra thought I think this is bad, because sometimes a `panic` only means the program don't know how to continue but doesn't necessarily means it cann't be recovered from last checkpoint.

fuyufjh:
After the dicussion, I'm a little wavering now. 

1. It introduces some extra complexity, not only in programing but also in analyzing a bug afterwards.
2. Sometimes it's hard to tell exactly whether a situation is non-recoverable or recoverable

xxchan:
> 1. It introduces some extra complexity, not only in programing but also in analyzing a bug afterwards.
> 2. Sometimes it's hard to tell exactly whether a situation is non-recoverable or recoverable

This sounds just the original reasons why we didn't do it..

hzxa21:
We haven't reach consesus during the meeting but many thoughts pop out.

## Goal
1. Let user aware of the happening of the errors
2. Maintain correct results of the MV/Sink when non-recoverable errors happen
3. Reduce the impact of non-recoverable errors for streaming and serving


## Meeting summary and some thoughts afterwards
- Goal `#1` can be achieved via reporting errors in grafana/cloud portal. We can also display the errors in a system table or display them on batch query if we want to draw attentions from the users.
- Goal `#2` (strict correctness - considering both MV and Sink) can be achieved via suspending the MV as well as all its downstream MVs but it hurts availablity of a subset of the streaming jobs.
- Goal `#3` can be achieved via
  - Not suspending any MVs when errors happen but just skipping the errors or filling in default value. **Availabilty is high but this hurts goal `#2`**.
  - Suspending only the affected MVs. **Availibility is downgraded but goal #2 can still be achieved**.
- Other thoughts brought up in the meeting:
  - Errors/panics caused by kernel bugs may be recoverable by upgrading the codes with the fixes and resume MVs from the lastest checkpoint.
  - We can provide an option to the user to choose whether he/she wants to suspend MV on non-recoverable errors but what the default behavior should be is still debatable.
  - We also talk about whether it is possible to keep the temporal join MV running if only its inner side MV is suspended. IMO, it is doable but complicated. We need atomic rename for MV so that the inner side MV can be recovered.
  - Not all panics are non-recoverable. Sometimes developer may just panic on meeting recoverble errors without thinking all through. In such cases, recovery may help.

In fact, goal `#2` contradicts with goal `#3` and user may pay different attentions depending on their use cases or development stages:
- In test environment or use cases that cannot tolerate incorrect results (e.g. sinking a wrong signal to external system may not be acceptable), use may want to catch the non-recoverable errors and fix them immdieately before bad things happen. In this case, correnectness is preferred over availibility.
- In production or use cases that can tolerate incorrect results, user may want its service to be always up. In this case, MV suspension may be a concern since availibility is preferred over strict correctness.

hzxa21:
> In production or use cases that can tolerate incorrect results, user may want its service to be always up. In this case, MV suspension may be a concern since availibility is preferred over strict correctness.

This is the key concern of this RFC. However, in the current implementation, even though incorrect result is acceptable by the user, not all erros can be skipped. Let's assume all the errors can be skipped are recoverable errors. I suggest we constrain our discussion and focus on the following two issues:

## How to identify the errors that cannot be skipped?

If we cannot identify these errors, we cannot do anything about it. The main debates are about how to handle panic.

The only feasible option seems to be letting developer carefully identify non-recoverable panic and report it as an non-recoverable error because as mentioned in the [comment](https://github.com/risingwavelabs/rfcs/pull/54#discussion_r1126021616), we can hardly know the origin actor of a panic.

Example: we already did the conversion from panic to error for the `MemTableError::InconsistentOperation` and this error is non-recoverable: https://github.com/risingwavelabs/risingwave/blob/f5e41a190b284b26a83b7740af77c842d5e35837/src/storage/src/mem_table.rs#L99

## What should we do when encountering errors that cannot be skipped?

Possible options:
1. Trigger recovery over and over again, hoping that recovery can fix the errors.
2. Pause all MVs and tell users to intervene. From user points of view, this is equivalent to option 1 if recovery cannot fix the errors (endless recovery == pausing all MVs).
3. Pause some MVs and tell users to intervene. This is the proposal from this RFC.

PR #55, opened 2023-02-27
RFC: decouple sink and checkpoint

wenym1:
[A rendered link](https://github.com/risingwavelabs/rfcs/blob/yiming/0055/rfcs/0055-decouple-sink-and-checkpoint.md)


wenym1:
We had a meeting last Friday. Here is some summaries:

- For a quick and simple implementation, we can implement a log store via kv interface first.
- Sinks can asynchronously consume and commit the data. Only the unconsumed data will be flushed to persistent storage for consistency of checkpoint. Sink can report consume progress to help deleting the consumed data.

xxchan:
Is this finished (and can be merged)?

fuyufjh (approved):

PR #56, opened 2023-02-27
RFC: Block Format V2

Li0k:


Li0k:
Summary
1. Introduce different encoding for flexible length types.
2. Remove the epoch of the key that is lower than the watermark on the bottom level, and remove the redundant encoding of table_id in key (in storage).
3. Rely on statistical information to determine whether to introduce dictionary encoding for epoch.

xxchan (approved):
Completed at https://github.com/risingwavelabs/risingwave/issues/8252

PR #59, opened 2023-04-10
RFC: DML Atomicity

chenzl25:


rfcs/0059-dml-atomicity.md
```diff
@@ -0,0 +1,47 @@
...
+
+- Send `Begin(txn_id)` to `DMLExecutor`, where `txn_id` is a transaction ID that can be generated using the same algorithm as `row_id`.
+- Send `Chunk(txn_id)` to `DMLExecutor`.
+- Send `End(txn_id)` to `DMLExecutor`.
+
+(Note that all of the above messages must be sent to the same `DMLExecutor` e.g. using `hash(txn_id)` .)
+
+
+As for DMLExecutor, maintain two internal states: 
+1. create table redo_log(txn_id, ops, column_1, …, column_N). ops and columns are exactly the same meanings as the stream chunk.
```
st1page, line 32: it looks very like the logStore's implementation in https://github.com/risingwavelabs/rfcs/pull/55

rfcs/0059-dml-atomicity.md
```diff
@@ -0,0 +1,47 @@
...
+2. create table commit_log(txn_id, state). The value of the state column could be uncommitted or committed
+
+The DMLExecutor should follow this contract:
+- `Begin(txn_id)` : insert txn_id with the uncommitted state into the commit log.
+- `Chunk(txn_id)` : insert chunks into the corresponding txn_id redo log.
+- `End(txn_id)` : Update the txn_id from an uncommitted to a committed state in the commit log. Then, apply the records from the redo log to the downstream. Applying records involves deleting them and sending them downstream. After applying the records, delete the txn_id from the commit log.
```
st1page, line 38: - must the redo log be applied in one epoch(or checkpoint)?
- must we blocking the batchDML executor before we finish the checkpoint?

rfcs/0059-dml-atomicity.md
```diff
@@ -0,0 +1,47 @@
...
+- Send `Begin(txn_id)` to `DMLExecutor`, where `txn_id` is a transaction ID that can be generated using the same algorithm as `row_id`.
+- Send `Chunk(txn_id)` to `DMLExecutor`.
+- Send `End(txn_id)` to `DMLExecutor`.
+
+(Note that all of the above messages must be sent to the same `DMLExecutor` e.g. using `hash(txn_id)` .)
+
+
+As for DMLExecutor, maintain two internal states: 
+1. create table redo_log(txn_id, ops, column_1, …, column_N). ops and columns are exactly the same meanings as the stream chunk.
+2. create table commit_log(txn_id, state). The value of the state column could be uncommitted or committed
```
st1page, line 33: should we delete all the uncommitted txn because we do not know if there are some records lost?

rfcs/0059-dml-atomicity.md
chenzl25: > must the redo log be applied in one epoch(or checkpoint)?

No. I think DMLs shouldn't block any barrier because our streaming engine relies heavily on barriers. And the eventual consistency seems good enough for DMLs.

> must we blocking the batchDML executor before we finish the checkpoint?

No. We can still follow the current behavior, the batchDML can response after `End(txn_id)` is sent.

rfcs/0059-dml-atomicity.md
chenzl25: Yes. According to the contract,  after recovery,  if it is found that the state of the commit log is `uncommitted`, it means this transaction hasn't passed the commit point, so we have to delete all the records.

rfcs/0059-dml-atomicity.md
BugenZhao, line 38: When will we notify the batch executors that the DML has successfully finished? Since the checkpointing is asynchronous, is it possible that the records are not persisted (durable) even if users get "success"? Is it tolerable for now?

BugenZhao (commented):
I begin to miss the old days when we treat the `risingwave-batch` as an external system and make it a special "connector". In this way, we may be able to resolve DML atomicity along with CDC transaction. 🥺

rfcs/0059-dml-atomicity.md
st1page: what is the primary key of the table? maybe the following operation on the same pk need to overwrite the previous and the DMLExecutor will actually generate a "upsert stream". But no need to worry, the DMLExecutor will always follows a Materialize executor with pk check

rfcs/0059-dml-atomicity.md
chenzl25: It is tolerable for now because we follow previous behaviors, but it will definitively confuse those users from OLTP databases. Our checkpoint interval currently is about 1s which is too long for a simple DML statement.

st1page: Just want to know if we can have the design to let the user know if the DML is commit successful or field :thinking:

BugenZhao: Agree. After this RFC, atomicity can be achieved with DML statements, and I think durability can be still achieved with `FLUSH`.

st1page: > atomicity can be achieved with DML statements, and I think durability can be still achieved with FLUSH. 

We can only achieve that if the DML executor and FLUSH can be execute atomically(both sucess or both fail)

ZENOTME: > Update the txn_id from an uncommitted to a committed state in the commit log. Then, apply the records from the redo log to the downstream.

Do we need guarantee that the redo log be persisted before apply the records?

I'm confused because according to follow Q&A, for my understand, you means that the batchDML will return response user if it get End(txn_id). But in this time it will not guarantee persisted the redo log and apply the record, right? So in this case if a crash occurs now, it will caused that the user get a response but the insert will failed, right?

>> must we blocking the batchDML executor before we finish the checkpoint?
No. We can still follow the current behavior, the batchDML can response after End(txn_id) is sent.

>> When will we notify the batch executors that the DML has successfully finished? Since the checkpointing is asynchronous, is it possible that the records are not persisted (durable) even if users get "success"? 
It is tolerable for now because we follow previous behaviors, but it will definitively confuse those users from OLTP databases. Our checkpoint interval currently is about 1s which is too long for a simple DML statement.

chenzl25: > Do we need guarantee that the redo log be persisted before apply the records?
> 
> I'm confused because according to follow Q&A, for my understand, you means that the batchDML will return response user if it get End(txn_id). But in this time it will not guarantee persisted the redo log and apply the record, right? So in this case if a crash occurs now, it will caused that the user get a response but the insert will failed, right?

Yes, in the case you mentioned, it is possible that users would get a response but actually the insert failed. We don't provide durability right now 🥵 .

rfcs/0059-dml-atomicity.md
chenzl25: Good question. I think the primary key of the `redo_log` should be `txn_id` plus the dml table's pk, so the row id generator should also be merged into the `DMLExecutor `.

rfcs/0059-dml-atomicity.md
chenzl25: This design relies on the checkpoint to commit DML statements, unless we can emit a checkpoint barrier right after a DML statement, but it seems expensive.

hzxa21: > > must the redo log be applied in one epoch(or checkpoint)?
> 
> No. I think DMLs shouldn't block any barrier because our streaming engine relies heavily on barriers. And the eventual consistency seems good enough for DMLs.

Doesn't it mean that batch query can see partial update states?

rfcs/0059-dml-atomicity.md
hzxa21: > it looks very like the logStore's implementation in #55

Indeed. cc @wenym1

rfcs/0059-dml-atomicity.md
BugenZhao: +1, but it sounds like the property of "Isolation"?

rfcs/0059-dml-atomicity.md
```diff
@@ -0,0 +1,55 @@
...
+
+
+![](images/0059-dml-atomicity/recovery.svg)
+
+As we can see, DML statements write their chunk into the redo log before actually applying them to the downstream. This is done so that we can ensure atomicity. If a recovery event occurs, we need to check the commit state to decide whether to apply or discard the data in the redo log. A `committed` state means we need to apply the remaining records in the corresponding redo log to the downstream, while an `uncommitted` state means we need to clean up the state of the corresponding redo log. We can perform this cleanup job after the `DMLExecutor` receives the first barrier. In summary, we should scan the commit log and clean up the redo log after recovery.
+
+## Durability
+
+Durability isn't guaranteed in this RFC. DMLs are considered finished after sending `End(txn_id)` to the `DMLExecutor`. The reason why we don't guarantee right now is that the current design relies on the checkpoint to commit DML statements, unless we can emit a checkpoint barrier right after a DML statement to force a checkpoint, but it seems expensive. Might be DMLs plus `Flush` is a more pratical solution to the durability.
```
BugenZhao, line 47: +1. We can send the `curr_epoch` after the `End` back to the batch executors, and only after they find this epoch is checkpointed successfully we can ensure the durability. This is somehow an inline `FLUSH`.

fuyufjh (approved):

PR #64, opened 2023-07-06
RFC: New Source DDL

tabVersion:


rfcs/0064-new-source-ddl.md
```diff
@@ -0,0 +1,303 @@
...
+    "c" int,
+    "d" int,
+    primary key ( a, b ) ) -- c is missing
+    format upsert encode [ JSON | AVRO | PROTOBUF ]
+    KEY ENCODE [ JSON | AVRO | PROTOBUF ] DATATYPE ( struct<a int, b int, c int> )
+    with (
+        'connector' = '...',
+        ...
+    );
+```
```
st1page, line 249: We can use generated column to express it.
```sql
create table [if not exist] source_1 (
    c int,
    d int, 
    a int as _key.a,
    b int as _key.b
    primary key ( a, b ) ) -- c is missing
    format upsert encode JSON 
    KEY ENCODE JSON DATATYPE (struct<a int, b int, c int>)
    with (
        'connector' = '...',
        ...
    );
```

```sql
/* schema registry*/
create table [if not exist] source_1 ( primary key ( _key.a as a, _key.b as b ) )
    format upsert encode PROTOBUF ( schema.location = "file://...") 
    KEY ENCODE JSON DATATYPE (struct<a int, b int, c int> )
    with (
        'connector' = '...',
        ...
    );
```

st1page:
progress update: 
- In release 1.0, we have changed the old `ROW FORMAT` syntax into new `FORMAT ... ENCODE ...` syntax
- we will not introduce KEY clause currently until we meet the real-world requirements, currently the default behavior that use a hidden column `_rw_key Bytea` is enough https://github.com/risingwavelabs/risingwave/pull/11125

fuyufjh (approved):

PR #68, opened 2023-07-31
add error_record_table

TennyZhuang:


rfcs/0068-error-record-table.md
```diff
@@ -0,0 +1,85 @@
...
+authors:
+  - "TennyZhuang"
+start_date: "2023/07/31"
+---
+
+# Error Record Table
+
+## Summary
+
+Our current streaming engine does not help users to discover, debug, and handle errors well. When user met an data record error, they can only find a log record like ``ExprError: Parse error: expected `,` or `]` at line 1 column 10 (ProjectExecutor: fragment_id=19007)``.
```
xxchan, line 12: FWIW, currently we also have error reporting to Prometheus (https://github.com/risingwavelabs/risingwave/issues/7824), but it do not contains original data (seemly for the same reason described in the Alternatives section)

xxchan: @neverchanje  proposed similar ideas in https://github.com/risingwavelabs/risingwave/issues/7803

> But when it comes to production deployment, we will need to provide an option that allows users to persist lost messages somewhere (maybe in Hummock or a log store) so that they can search for what reason and what data were lost (maybe through ElasticSearch).

Closed as "we are now able to find the errors from Grafana", but we haven't supported "allow users to persist lost messages" yet.

rfcs/0068-error-record-table.md
```diff
@@ -0,0 +1,96 @@
...
+
+Same as other internal tables while suffixed by `error_{seq}`.
+
+### Schema
+
+The schema of ERT should have the same fields as their input, with several extra columns:
+
+1. `id bigint`: The ID can be generated by the similar method like `row_id` (vnode + local monotical ID).
+2. `error_reason varchar`: A human-readable error message.
+
```
kwannoel, line 51: Do we also have a column for sql? Or store it in the `error_reason` col?

rfcs/0068-error-record-table.md
```diff
@@ -0,0 +1,96 @@
...
+1. Expression errors (division by zero, cast fail, json extraction fail)
+2. UDF errors (timeout, invalid records, user errors)
+3. Source error (paring failed)
+
+## Design
+
+### Creating
+
+The ERTs are automatically created as internal tables when an operator is created. In most cases, an operator will have n ERTs, where n corresponds to the number of inputs it has.
```
fuyufjh, line 37: So how can a user get all errors happened in a MV?

rfcs/0068-error-record-table.md
kwannoel: IMO we may store error ID instead of varchar here?

kwannoel: Then store error_reasons in a system table.

rfcs/0068-error-record-table.md
BugenZhao, line 37: Do we have any implementation details in this RFC? For example, how will this change the interface of expression evaluation? Will we have to pass the new `StateTable` for error records everywhere?

rfcs/0068-error-record-table.md
kwannoel: > Do we also have a column for the sql? Or store it in the error_reason col?
Store the mview ID.

rfcs/0068-error-record-table.md
fuyufjh: Also, I think it sounds too heavy to keep one error table for each operator. I tend to keep one single error table for the whole database.

rfcs/0068-error-record-table.md
kwannoel: > IMO we may store error ID instead of varchar here?

Yes we can, but have to deal with backwards compat issues, in case users construct their own mapping logic using the error ids.

rfcs/0068-error-record-table.md
BugenZhao: +1. Note that this makes stateless operators like `Project` and `Filter` contain state tables as well. We even have to handle stuff related to distributed execution for them, like scaling. From this perspective, it seems too invasive to me.

rfcs/0068-error-record-table.md
```diff
@@ -0,0 +1,96 @@
...
+1. The data record may be too large to record, e.g. several tens of KB.
+2. Errors may occur continuously, causing the log system to fill up quickly.
+
+From the viewpoint of users, many of them do not consider logs as being reliable. In our default error handling approach, we do not halt the stream when encountering stream errors, which may lead users to anticipate greater reliability in those logs.
+
+## Future possibilities
+
+### Data correction
+
+ERT could potentially be used to correct data, for example, users could clean up the data within ERT and then reimport it into the source.
```
BugenZhao, line 79: The record in ERT could be an intermediate result of arbitrary operators in the plan, so it might be difficult for users to recognize or understand the values (unless he is an expert of RisingWave and knows the execution details). So I'm afraid users are not always able to operate on these "discarded" records, not to mention to correct them, except for some special ones like the errors for source parsing or type casting.

fuyufjh:
After some extra thinking, I would like to propose to use a single error_table rather than one error_table per operator.

As mentioned in the RFC, our targets include

> 1. We can ensure that our storage engine can handle the volume of erroneous data, as it is of the same magnitude as the source.
> 2. We can ensure the error records are durable.
> 3. Users can view the error records directly over psql.
> 4. Users can reproduce the error easily by the similar SQL.

In short, turning error_table per operator into a single error_table will basically enhance 3 a lot and slightly impair 4. Additionally, it avoids introducing lots of error_tables, which will be confusing for most people that don't really care errors; and keeps backward compatibility.

I'll explain these pros and cons one by one.

**"Enhance 3: Users can view the error records directly over psql"**

In the single-table approach, the users only need to query the `rw_error_table` and he will get all errors happened recently. This is quite intuitive to anyone.

Instead, in the error_table-per-operator fashion, we might need to provide an extra 'view' for that purpose. Note that this is not an actual 'view' because when a new materialized view appears, the view needs to be updated accordingly, which is somehow dirty to me.

**"Slightly impair 4: Users can reproduce the error easily by the similar SQL."** 

In most cases, users don't need to really reproduce it in RisingWave (except it's RW's bug, but that's another story). For example, as a user, if I encountered a "JSON Parse" exception, It would be enough to tell me the **record id** and I can check the JSON in upstream by myself. Furthermore, I'll try to fix that data and optionally emit the record into Kafka again to let RisingWave consume it. Reproducing it in RisingWave side doesn't help me a lot.

**"It avoids introducing lots of error_tables"**

This is ugly both from user side and for us.

From user side, our `SHOW ALL TABLES` command is supposed to show all internal state tables. It just looks confusing if more than half of them are `xxx_error_record_tables`. 

For us, we now assume a table is one of these 3 types: tables, materialized views and streaming states. The `xxx_error_record_tables` will be a new kind and creates lots of related things: compaction group, monitoring metrics, etc. 

**"Keeps backward compatibility"**

Because their will be only one error_table, we can simply hard-code this table in system catalog. That means that we don't need to change the plan node of existing fragments, limiting the changes only in operator's code implementation.

---------------

Finally, since we have discussed that we will write into "error_record_tables" with blind write and singleton distribution (which means the writers aren't aware of any distribution, simply wirte with a random unique ID), both approach have similar complexity in terms of the core implementation (i.e. write-path).

kwannoel:
> "Enhance 3: Users can view the error records directly over psql"
>
> In the single-table approach, the users only need to query the rw_error_table and he will get all errors happened recently. This is quite intuitive to anyone.
>
> Instead, in the error_table-per-operator fashion, we might need to provide an extra 'view' for that purpose. Note that this is not an actual 'view' because when a new materialized view appears, the view needs to be updated accordingly, which is somehow dirty to me.

For 3. I didn't quite get it. Why don't we just `CREATE VIEW` to simulate the `rw_error_table` which contains all recent errors. Rather than `CREATE MATERIALIZED VIEW`. This runs the computation adhoc, rather than incrementally?

> Finally, since we have discussed that we will write into "error_record_tables" with blind write and singleton distribution (which means the writers aren't aware of any distribution, simply write with a random unique ID), both approach have similar complexity in terms of the core implementation (i.e. write-path).

How do we synchronize access to the state table? Or we use uuid here to synchronize, instead of row_id? If yes that makes sense 👍 

Also I'm assuming the error record will be stored as jsonb? Is our `jsonb` implementation still fixed size, so it could lead to data loss?
I guess  another advantage of ERT per executor is that we won't encounter this?

rfcs/0068-error-record-table.md
kwannoel: > +1. This inspires me that it makes stateless operators like Project and Filter contain state tables as well. We even have to handle stuff related to distributed execution for them, like scaling or plan evolution. From this perspective, it seems too invasive to me.

For scaling I think it shouldn't be a problem if we follow's Eric's approach of uuid + singleton distribution.

fuyufjh:
> > Instead, in the error_table-per-operator fashion, we might need to provide an extra 'view' for that purpose. Note that this is not an actual 'view' because when a new materialized view appears, the view needs to be updated accordingly, which is somehow dirty to me.
>
> For 3. I didn't quite get it. Why don't we just CREATE VIEW to simulate the rw_error_table which contains all recent errors. Rather than CREATE MATERIALIZED VIEW. This runs the computation adhoc, rather than incrementally?

No, I'm not using MATERIALIZED VIEW. 

A "View" in database is defined by a SQL query: `create view <view_name> as <query>`. However, here we have multiple error_tables with different schema:

```
table name                   schema
error_table_of_operator_a    col1 (int)  col2 (varchar)   error_message (varchar)
error_table_of_operator_b    col1 (int)  col2 (decimal) col3 (varchar)  error_message (varchar)
error_table_of_operator_c    col1 (varchar)  col2 (varchar) col3 (int)  error_message (varchar)
```

If we want to query from all error tables, then the query would be like:

```sql
/* note the columns are lost because they can't be merged into one result set */
select error_message from error_table_of_operator_a
union all
select error_message from error_table_of_operator_b
union all
select error_message from error_table_of_operator_c
```

Furthermore, if you materialized view is created or dropped, you need to update the view accordingly

```sql
/* note the columns are lost because they can't be merged into one result set */
select error_message from error_table_of_operator_a
union all
select error_message from error_table_of_operator_b
union all
select error_message from error_table_of_operator_c
union all
select error_message from error_table_of_operator_d -- d is a new operator
```

That's why I said it's kind of dirty - We need to hook the DDL processes. Or, alternatively, give up the SQL approach and provide a hard-code access way. This is also dirty because we need to construct such a query plan in hard-code way, instead of running normal optimizer process. 😇

fuyufjh:
> How do we synchronize access to the state table? Or we use uuid here to synchronize, instead of row_id? If yes that makes sense 👍

Yes, uuid should be the most practical solution.

> Also I'm assuming the error record will be stored as jsonb? Is our jsonb implementation still fixed size, so it could lead to data loss?

Yes, JSONB is one of the option. Technically there is no reason of data loss. Actually I don't really care whether the record is kept losslessly, I think it's okay to truncate records to limit the size, because the error is only supposed for human users to read.

> I guess another advantage of ERT per executor is that we won't encounter this?

True.

PR #69, opened 2023-08-02
WIP RFC: lambda expression

st1page:
use lambda expression to enhance array's complex processing

### DIFF
--- a/rfcs/xxxx-lambda-expression
+++ b/rfcs/xxxx-lambda-expression
@@ -105,40 +105,79 @@ Btw, It often can be used to express UDAF too.
 
 ## Investigation
 
-Simple array expressions
+ Array expressions support. (All of them supports the `array_size/cardinality`, `unnest/flatten/explode` which is not listed in the table.)
 
-| array_sum | array_max/min | array_sort | array_zip | array_zip |
-|  ----  | ----  |
-|   |  |
-|   |  |
+|            | array_cnt | array_sum | array_cum_sum | array_avg |  array_max/min   | array_sort | array_distinct | array_zip | Unnest<br>with idx |
+| ---------- | :-------: | :-------: | :-----------: | :-------: | :--------------: | :--------: | :------------: | :-------: | :----------------: |
+| Databricks |           |           |               |           |        Y         |     Y      |       Y        |     Y     |         Y          |
+| Trino      |           |           |               |           |        Y         |     Y      |       Y        |     Y     |         Y          |
+| PrestoDB   |     Y     |     Y     |       Y       |     Y     |        Y         |     Y      |       Y        |     Y     |         Y          |
+| Starrocks  |     Y     |     Y     |       Y       |     Y     |        Y         |     Y      |       Y        |     Y     |                    |
+| Doris      |     Y     |     Y     |       Y       |     Y     |        Y         |     Y      |       Y        |     Y     |                    |
+| ClickHouse |     Y     |     Y     |       Y       |     Y     |        Y         |     Y      |       Y        |     Y     |                    |
+| DuckDB     |     Y     |     Y     |       Y       |     Y     |        Y         |     Y      |       Y        |           |                    |
+| Vertica    |     Y     |     Y     |               |     Y     |        Y         |            |                |           |         Y          |
+| FlinkSQL   |           |           |               |     Y     | only MAX, no MIN |            |                |           |                    |
+| Snowflake  |           |           |               |           |                  |            |       Y        |           |         Y          |
 
-array expressions with lambda expression paramter
 
-### PrestoDB
 
-PrestoDB supports many array aggregate expressions <https://trino.io/docs/current/functions/array.html?highlight=array#array_max>. Also it implement both scalar and aggregation lambda expression on array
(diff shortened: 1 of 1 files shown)


rfcs/0069-lambda-expression.md
```diff
@@ -0,0 +1,195 @@
...
+
+It is a kind of SQL **scalar expression** which accept a array type and return the array's aggregation result, such as `array_sum`, `array_max`. It can solve some part of the issue but can not handle the complex nested datatype in the array.
+
+```SQL
+select *, array_sum(arr_i) from t;
+```
+
+### level 2: scalar lambda expression
+
+Implement lambda expression function to make user can define there `transform` logic for array's each elements.
```
xiangjinwu, line 83: https://github.com/risingwavelabs/risingwave/issues/11123
Another part of this solution is to implement `transform` and `reduce` functions, whose inputs contain both data column and the lambda expression.

chenzl25:
I find that we can unnest this kind of subquery in some way. Here is a PR on how to transpose Apply with  ProjectSet. https://github.com/risingwavelabs/risingwave/pull/11390

st1page:
> I find that we can unnest this kind of subquery in some way. Here is a PR on how to transpose Apply with ProjectSet. https://github.com/risingwavelabs/risingwave/pull/11390.
If we support the following two optimizer rules later, we can resolve this issue.

LGTM, but still not the best because it introduce unnessary Join operator 🤔

chenzl25:
> > I find that we can unnest this kind of subquery in some way. Here is a PR on how to transpose Apply with ProjectSet. [risingwavelabs/risingwave#11390](https://github.com/risingwavelabs/risingwave/pull/11390).
> > If we support the following two optimizer rules later, we can resolve this issue.
> 
> LGTM, but still not the best because it introduce unnessary Join operator 🤔

True. It could be a temporal solution. From a performance perspective, this RFC could do better.

rfcs/0069-lambda-expression.md
```diff
@@ -0,0 +1,195 @@
+---
+feature: my_excited_feature
```
TennyZhuang, line 2:
```suggestion
feature: lambda_expression
```

rfcs/0069-lambda-expression.md
```diff
@@ -0,0 +1,195 @@
...
+start_date: "2023/08/01"
+---
+
+# Lambda expression
+
+## Background
+
+To better support user's semi-structure data, We have supported `Array` and `Struct` types compatible with PG. But in some user's requirement, We found we can not support them easily.
+
+### Induction
```
CAJan93, line 14:
```suggestion
### Introduction
```

CAJan93: Sorry. Tiny nit only. This idea really sounds good to me. Hope we will go for it :D

TennyZhuang:
The latest progress:

## The lambda syntax

In this RFC, we proposed using the arrow symbol “->” as the syntax for lambda expressions, such as `x -> x * 2`, since it's widely supported by different systems. However, this symbol conflicts with JsonAccess and creates significant ambiguity.

A bad case: how should `f(ARRAY[1,2,3], x -> 'y' or z)` be parsed? `(x -> 'y') or z` or `x -> ('y' or z)`? The AST will be much different after the parser phase.

We also proposed the `=>` symbol, which is used by `JavaScript` and also widely accepted by programmers. Unfortunately, it's still conflict with  a pg syntax, [Named notation](https://www.postgresql.org/docs/current/sql-syntax-calling-funcs.html#SQL-SYNTAX-CALLING-FUNCS-NAMED).

```sql
-- Lambda function
SELECT f(ARRAY[1,2,3], x => x * 2);
-- Named notation
SELECT abs(a => -1)
-- ???
SELECT f(apply => (x => x * 2), array => ARRAY[1,2,3]);
```

There are definitely many compiler techniques that can solve the problem of ambiguity in the above syntax. After all, it is much simpler compared to C++ syntax. However, as an experimental feature, we do not want to introduce very complex refactors into the parser for it. We have decided to look for a syntax that is not ambiguous and relatively simple to support this feature. **The final choice is a Rust-like syntax `|x| x * 2`**.

In the current pg 15, there is no built-in operator that allows the prefix `|` operator. It may conflict with custom operators, but currently we do not support the feature. Even if we support it, it 's still acceptable to forbid “|”” as a unary operator.

wangrunji0408:
It is possible to introduce another macro to generate array scalar functions from aggregate definitions. It may look like:

```rust
#[aggregate("sum(int64) -> int64")]                // original aggregate
#[array_function("array_sum(int64[]) -> int64")]   // the new macro
fn sum(state: i64, input: i64) -> i64 {
    state + input
}
```

This approach requires no extra code for each function, instead it puts all complexity into the macro.

By contrast, manually implement each `array_*` function is a bit tedious but trivial:

```rust
#[function("array_sum(int64[]) -> int64")]
fn array_sum(array: ListRef<'_>) -> i64 {
    array.iter().map(...).sum()
}
```

So I would also prefer manual implementing each array function.

xiangjinwu:
Just ran into this:
https://www.postgresql.org/docs/current/intarray.html

PR #70, opened 2023-08-02
RFC: User Defined Aggregate Functions

wangrunji0408:


rfcs/0000-user-defined-aggregate-functions.md
```diff
@@ -0,0 +1,269 @@
...
+
+# The aggregate function is defined as a class.
+# Specify the schema of the aggregate function in the `udaf` decorator.
+@udaf(input_types=['BIGINT', 'INT'], result_type='BIGINT')
+class WeightedAvg:
+    # Create an empty accumulator.
+    def create_accumulator(self) -> State:
+        accumulator = State()
+        accumulator.sum = 0
+        accumulator.count = 0
```
stdrc, line 40: In Python we can also do the following so that no need to manually initialize each field (if we meant to use this as a demo in document)

```py
class State:
    sum: int = 0
    count: int = 0
```

rfcs/0000-user-defined-aggregate-functions.md
```diff
@@ -0,0 +1,269 @@
...
+```sql
+create aggregate weighted_avg(value bigint, weight int) returns bigint
+as 'weighted_avg' using link 'http://localhost:8815';
+```
+
+The full syntax is:
+
+```sql
+CREATE [ OR REPLACE ] AGGREGATE name ( [ argname ] arg_data_type [ , ... ] )
+[ RETURNS return_data_type ] [ APPEND ONLY ]
```
stdrc, line 155: Can we use `result_data_type` here to maintain wording consistency? Or use `return_type` everywhere, just want consistency.

rfcs/0000-user-defined-aggregate-functions.md
```diff
@@ -0,0 +1,269 @@
...
+    deactivate CP
+```
+
+### State Storage
+
+The state of UDAF is managed by the compute node as a single encoded BYTEA value.
+
+Currently, each aggregate operator has a **result table** to store the aggregate result. For most of our built-in aggregate functions, they have the same output as their state, so the result table is actually being used as the state table. However, for general UDAFs, their state may not be the same as their output. Such functions are not supported for now.
+
+Therefore, we propose to **transform the result table into state table**. The content of the table remains the same for existing functions. But for new functions whose state is different from output, only the state is stored. The output can be computed from the state when needed.
```
stdrc, line 211: Maybe `intermediate state table` is clearer.

rfcs/0000-user-defined-aggregate-functions.md
```diff
@@ -0,0 +1,269 @@
...
+| 2     | U+   | 0       | 2    | 1    | true  |
+
+The new **state table** (derived from old result table) of the agg operator would be like:
+
+| Epoch | id   | sum(v0) | bool_and(v1)        | weighted_avg(v0, w0) | max(v0)*   |
+| ----- | ---- | ------- | ------------------- | -------------------- | ---------- |
+| 1     | 0    | sum = 1 | false = 1, true = 0 | encode(1,2) = b'XXX' | output = 1 |
+| 2     | 0    | sum = 2 | false = 0, true = 1 | encode(2,1) = b'YYY' | output = 2 |
+
+* For **append-only** aggregate functions (e.g. max, min, first, last, string_agg...), their states are all input values maintained in seperate "materialized input" tables. For backward compatibility, their values in the state table are still aggregate results.
```
stdrc, line 228: non-append-only?

rfcs/0000-user-defined-aggregate-functions.md
```diff
@@ -0,0 +1,269 @@
...
+class State:
+    sum: int
+    count: int
+
+# The aggregate function is defined as a class.
+# Specify the schema of the aggregate function in the `udaf` decorator.
+@udaf(input_types=['BIGINT', 'INT'], result_type='BIGINT')
+class WeightedAvg:
+    # Create an empty accumulator.
+    def create_accumulator(self) -> State:
```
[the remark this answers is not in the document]
wangrunji0408: Agree. 😁

rfcs/0000-user-defined-aggregate-functions.md
wangrunji0408: Non-append-only (retractable) functions are described above as a general case. This section is just a note for the special case.

rfcs/0000-user-defined-aggregate-functions.md
```diff
@@ -0,0 +1,266 @@
...
+
+The remaining clauses are defined the same as for `create function` statement.
+
+## Design
+
+### Dataflow
+
+User defined functions are running in a separate process called UDF server. The RisingWave kernel communicates with the UDF server through Arrow Flight RPC to exchange data.
+
+To avoid trouble in fault tolerance, the UDF server should be **stateless** even if the aggregate function is stateful. This means the aggregate state should be maintained by the kernel. However, exchanging the state in each RPC call (batch aggregation) is not efficient, especially when the state is large. On the other hand, kernel doesn't need to know the state or the aggregate result before a barrier arrives. Therefore, we **create a streaming RPC `doExchange` to call the aggregate function, and sync the state every time a barrier arrives**. When the connection to the UDF server is broken, the executor raises an error and may retry or pause the stream.
```
wangrunji0408, line 173: I found a fatal problem that the current design doesn't take group aggregation into consideration. We still have to send intermediate states for each group.

PR #72, opened 2023-08-18
RFC: Reusable Source Executor

fuyufjh:


rfcs/0061-reusable-source.md
```diff
@@ -0,0 +1,83 @@
...
+
+Recall that most source has more or less historical data. For message brokers including Kafka, Pulsar, Kinesis, etc, this can be configured with option `scan.startup.mode`. When `scan.startup.mode` is not set to `latest`, the consumer must read the historical data before consuming any new events. This process is implicitly done by the client library, and RisingWave is not aware of it.
+
+Let's assume that the Kafka client library is stupid and RisingWave needs to read the historical data by itself. Then, it turns out this procedure is quite similar to the backfilling procedure of tables. Recall that when a new materialized view is created, we introduce a `BackfillExecutor` to read the historical data. Here we can solve the problem with the same idea by dividing it into 2 phases: 1) backfill and 2) consume new events.
+
+![](images/0061-reusable-source/reusable-source.png)
+
+Here,
+
+- `KafkaSourceExecutor` is created on executing `CREATE SOURCE` statement, although a trivial optimization is doing nothing when there is no downstream operator attached.
```
StrikeW, line 27: It would be better to clarify the scope of this rfc. Does it mean all `CREATE SOURCE` will be instantiated a corresponding streaming job after this rfc has been done? Which is a fundamental changes to me.

fuyufjh: >  Does it mean all CREATE SOURCE will be instantiated a corresponding streaming job after this rfc has been done? 

Correct. 

> Which is a fundamental changes to me.

Yes, I agree, but mostly conceptually. The implementation change seems to be acceptable to me.

As stated in the text, currently, the behavior of source is somehow special. So it might be a problem that need to be fixed. I am not sure how users feel about it.

StrikeW: +1 to make `CREATE SOURCE` as a streaming job, which is a better way to share same configuration among multiple downstream mv jobs and may save resources. And we don't need to introduce a new `Changelog` stream job proposed in the [RFC: Share a change log stream for multiple CDC tables ](https://www.notion.so/risingwave-labs/RFC-Share-a-change-log-stream-for-multiple-CDC-tables-1f1581d76da2458ba6fafca0c5252226).

Btw I have some discussion with @BugenZhao about the implementation. Most of the work should be done on Meta to add a new kind of `Source` stream job or we can add a dumb sink node behind `Source` to treat it as a `Sink` stream job.

rfcs/0061-reusable-source.md
```diff
@@ -0,0 +1,83 @@
...
+
+So how to combine the backfill data with incoming data without duplicating or missing events?
+
+Since everything happens inside the `BackfillExecutor`, the implementation is not difficult. Here is the pseudo-code.
+
+```rust
+let source_stream = the stream from upstream SourceExecutor i.e. new events
+let backfill_stream = read from message brokers with user-defined scan_startup_mode) i.e. historical events
+
+// In fact, these should be per kafka partition
```
BugenZhao, line 46: It might be worth noting that split changes should also be applied to `Backfill` executors, and...
- we need to persist the backfill offsets in this case,
- aligning the distribution/scheduling of `Backfill` and `Source` could make life much easier.

rfcs/0061-reusable-source.md
```diff
@@ -0,0 +1,83 @@
...
+
+Note that our design only applies to sources, excluding the tables with connectors, which are already instantiated and shared between all materialized views. By the way, [RFC: CDC Source with Backfill](https://github.com/risingwavelabs/rfcs/pull/63) takes a similar idea by introducing a dedicated `CdcBackfillExecutor` operator.
+
+### Implementation for Message Brokers
+
+Events from Kafka should be processed according to their sequence in partition, no matter for historical events or new incoming events.
+
+So how to combine the backfill data with incoming data without duplicating or missing events?
+
+Since everything happens inside the `BackfillExecutor`, the implementation is not difficult. Here is the pseudo-code.
```
BugenZhao, line 40: So this seems to be another implementation besides MV-on-MV backfill and CDC backfill, since each of them seems to have a slightly different algorithm?

Hope that there's some abstraction for this new backfill executor, so that we don't have to introduce a physical executor every time we support a new source connector. 🥵

rfcs/0061-reusable-source.md
BugenZhao, line 27: I guess this "optimization" is necessary, as there're kinds of sources that do not support rewinding the historical data like `Nexmark` or `Datagen`. So it can be confusing if the source executor runs itself and loses some records in the downstream materialized views created later.

BugenZhao (commented):
Really love this idea! This makes "Source" in our system much more well-defined.

rfcs/0061-reusable-source.md
fuyufjh: Yeah, there is already pretty lots of duplication between MV-on-MV backfill and CDC backfill. I am not sure whether it's possible to DRY it. Anyway, either DRY or not is acceptable to me.

st1page (commented):
Love this idea. Some supplements:
1. For the watermark on the source, We need to add the WatermarkFilter in the source fragment during the `CREATE SOURCE` statement. And I think we do not need to support do the watermark logic in the backfill part. because 
    - we have not supported and do not have a design about the MV on MV's watermark in the backfill phase. https://github.com/risingwavelabs/risingwave/issues/8375 So we should assume the downstream can accept there could be no watermark for a long time.
    - Our watermark filter algorithm is non-deterministic among different splits. So even when we filter the out-date data in backfill. Backfills in different streaming queries are still inconsistency. 
2. Based on this design, we will have a global unique instance for the source. And we can easily insert some common logic on the source and show it to users naturally. E.g, an internal table includes the record filtered by watermark. An internal table that includes the when parsing from the connector.

BugenZhao:
I personally prefer putting one `WatermarkFilter` after each `Backfill`, as different materialized views indeed filter out different records considering TTL on external sources. This also matches the current behavior that sources instantiated in different streaming jobs has their own watermarks.

rfcs/0072-reusable-source.md
```diff
@@ -0,0 +1,83 @@
...
+
+Events from Kafka should be processed according to their sequence in partition, no matter for historical events or new incoming events.
+
+So how to combine the backfill data with incoming data without duplicating or missing events?
+
+Since everything happens inside the `BackfillExecutor`, the implementation is not difficult. Here is the pseudo-code.
+
+```rust
+let source_stream = the stream from upstream SourceExecutor i.e. new events
+let backfill_stream = read from message brokers with user-defined scan_startup_mode) i.e. historical events
```
liurenjie1024, line 44: Technically, mv defined on source with `scan_startup_mode=earliest` are different when mvs are created in different time since kafka will clean up messages based on some criterial.

rfcs/0072-reusable-source.md
```diff
@@ -0,0 +1,83 @@
...
+### Implementation for Message Brokers
+
+Events from Kafka should be processed according to their sequence in partition, no matter for historical events or new incoming events.
+
+So how to combine the backfill data with incoming data without duplicating or missing events?
+
+Since everything happens inside the `BackfillExecutor`, the implementation is not difficult. Here is the pseudo-code.
+
+```rust
+let source_stream = the stream from upstream SourceExecutor i.e. new events
```
liurenjie1024, line 43: IIUC, this will require source executor to materialize messages?

fuyufjh: No, it's streamed from Kafka Consumer.

rfcs/0072-reusable-source.md
fuyufjh: Sure. If you create multiple MVs on one source, there is no consistency between them, because their `backfill_stream` may have different set of events.

rfcs/0072-reusable-source.md
liurenjie1024: Hmm, then what's the benefit of this extra source executor? Just to align with other objects like sink, mv?

rfcs/0072-reusable-source.md
```diff
@@ -0,0 +1,83 @@
...
+let backfill_stream = read from message brokers with user-defined scan_startup_mode) i.e. historical events
+
+// In fact, these should be per kafka partition
+let backfill_offset = 0;
+let backfill_completed = false;
+
+next_event = select! {
```
liurenjie1024, line 50: Does this require back fill executor has same parallelism as source executor?

StrikeW: I am concerning that backfill a source may delay the barrier event. If we bias for `source_stream`, the backfill progress may not catch up with upstream and hurt freshness, if we bias for `backfill_stream`, the barrier will be queued in the source. So may be we need a customized round-robin `select!` here?

rfcs/0072-reusable-source.md
xxchan: I think currently for N stream jobs on a source, we have N consumers. With this design, we have 1+N consumers during backfilling, and only 1 after backfill is finished.

rfcs/0072-reusable-source.md
fuyufjh: Here the barrier should be passed directly (aka. "stealing")

This is same with our design of any other source - the barrier will always be injected whenever the barrier comes

rfcs/0072-reusable-source.md
```diff
@@ -0,0 +1,83 @@
...
+
+# Reusable Source Executor
+
+
+## Motivation
+
+Conceptually, a source is very different from a table or a sink in terms of whether it is instantiated. A source defined by a `CREATE SOURCE` statement is not an instance, it's just a set of metadata stored in the catalog until it's referenced by a materialized view, at which point a `SourceExecutor` is instantiated.
+
+As a result, when creating multiple materialized views on top of one source, the `SourceExecutor` is not reused between them. It leads to multiple times of resource utilization and multiple consuming offsets, metrics, etc. Furthermore, it becomes a bug if the source definition uses some exclusive property, such as the consumer group of Kafka source.
+
```
kwannoel, line 16: We should include @xxchan's comment in this RFC: https://github.com/risingwavelabs/rfcs/pull/72/files#r1388947714

> I think currently for N stream jobs on a source, we have N consumers. With this design, we have 1+N consumers during backfilling, and only 1 in steady-state after backfill is finished.

It makes it a lot clearer.
```suggestion
Consider that now, for N stream jobs on a source, we have N consumers. With this design, we have 1+N consumers during backfilling (1 source, N backfill), and only source executor in steady-state after backfill is finished.
```

xxchan: I will update it to elaborate more the possible benefits.

PR #75, opened 2023-09-05
add expr-opaque-context

TennyZhuang:


rfcs/0075-expr-opaque-context.md
```diff
@@ -0,0 +1,161 @@
...
+
+The pattern was widely used in golang. `context` is a part of std lib.
+
+There are also a hook [`useContext`](https://react.dev/reference/react/useContext) in react, which is excatly the same pattern as the RFC proposed.
+
+## Alternatives
+
+### task_local
+
+tokio's [`task_local` macro](https://docs.rs/tokio/latest/tokio/macro.task_local.html) can achieve the same result even simpler, and the only cost is the extensive use of global variables.
```
BugenZhao, line 157: `task_local` is not that "global" as the calling stack or the scope of evaluating an expression tree is quite clear. I would vote for this approach as it introduces minimal changes to the codebase.

wangrunji0408: +1 for this approach. Adding `context` to the `eval` interface will make things more complicated and prevent easy use for most of the pure functions. Every time we call `eval`, we have to find out a context to pass in...

TennyZhuang: You can even pass a () as context here. I don’t think it’s a heavy problem.

The key concept is pass it explicitly.

TennyZhuang: We can also add a eval_with_empty_context.

It’s really simple.

TennyZhuang: In fact, all `eval` needs a context and I'd prefer pass it explicitly instead of the one from unknown scope.

wangrunji0408: Yes, my point is explicit context will put more burden on developers and make it error-prone. It's possible that developers forget to pass the context down through. This happened in our batch module, where we forgot to pass the shutdown channel which results in batch query not cancellable.

Another example is the `Future` API. Even if it has an explicit context in the `poll` method, most of the time the context is passed implicitly via `await`. 🤔

TennyZhuang: I don't want to depend on the call stack heavily. If the code path changed, the code may be broken without any warnings.

BugenZhao: Well, I guess we can have an `eval_with_context` on the root and set the task-local variable there. In this way, theses local keys can be private in the `expr` crate and we can find where the context is set by listing all references of this function.

wangrunji0408: Another solution: is it possible to put the context into the expression itself? I've discussed it with @shanicky yesterday.

TennyZhuang: > Well, I guess we can have an eval_with_context on the root and set the task-local variable there

If you've already got the context at the expr root, it's really easy to just pass them recursively. It also gives you opportunity to override some variables with tracking.

The most of the work is, how to pass these variables from frontend to executors then expressions.

TennyZhuang: > Another solution: is it possible to put the context into the expressions themselves? I've discussed it with @shanicky yesterday.

The lifetime of the context will be longer, it's not friendly for MetaCache, which is actually a ReadGuard and hold the read lock.

I also fully discussed with @fuyufjh last week.

TennyZhuang: > Well, I guess we can have an eval_with_context on the root and set the task-local variable there.

Another issue is that our original intention was to make the expression module unaware of the values and types of these variables, but you completely deviated from this purpose.

wangrunji0408: > > Well, I guess we can have an eval_with_context on the root and set the task-local variable there.
> 
> Another issue is that our original intention was to make the expression module unaware of the values and types of these variables (likely defined in frontend), but you completely deviated from this purpose.

In fact, these functions can be defined in the frontend module. The `#[function]` macro will register it into a global registry at start time. 🤡

BugenZhao: > Another issue is that our original intention was to make the expression module unaware of the values and types of these variables (likely defined in frontend), but you completely deviated from this purpose.

This sounds impossible. Although the context struct could be opaque, the type of each field must be accessible by the `expr` crate for expressions to use them.

wangrunji0408: > > Another issue is that our original intention was to make the expression module unaware of the values and types of these variables (likely defined in frontend), but you completely deviated from this purpose.
> 
> This sounds impossible. Although the context struct could be opaque, the type of each field must be accessible by the `expr` crate for expressions to use them. Given that every field's type is known, whether the context itself should be concrete or opaque won't be a big deal.

Yes, so it seems that we **should and only can** define these function in the frontend.

TennyZhuang: Yes, but we need to move the expression construction out of the expr crate.

fuyufjh: Besides, from my understanding, @TennyZhuang's point to introduce `provide` instead of using a simple `struct ExprContext` is he wants to make it more flexible. For example, if `BackendContext` exists, then the `eval()` function can get it; otherwise, it gets `None`.  So if you are sure that no context is needed somewhere (I don't know what can be the case... but just saying), it's possible to use an empty context.

wangrunji0408: > Yes, but we need to move the expression construction out of the expr crate.

Not necessary? We can pass the opaque context into the constructor function.

wangrunji0408: > I think evaluating expression doesn't happen everywhere.

It does happen everywhere in unit tests. And we have to provide an empty context for each of them. However, I think providing a `eval_with_default_context` method is acceptable.

fuyufjh: > It does happen everywhere in unit tests. 

Well. Indeed. 😄

If it costs too many changes to replace `eval` to `eval_with_default_context`, scoped task_local might look a little bit better. Either way LGTM.

By the way, it's worth mentioning that using `provide` or not is orthogonal to whether to put context in function call arguments or `task_local`: you may also put the `Context: Provider` as a task-local variable.

fuyufjh (approved):

PR #77, opened 2023-11-14
feat: Add shuffle for iceberg sink

liurenjie1024:


rfcs/0077-iceberg-sink-shuffle.md
```diff
@@ -0,0 +1,72 @@
...
+    B -->|IcebergDispatcher| C[IcebergSink]
+```
+
+In this approach we need to add an extra `IcebergDispatcher` to dispatcher executor. The `IcebergDispatcher` will calculate the partition value of each record and do hash shuffle according to the partition value.
+
+I prefer approach 1 since it's more extensible and does not change too much current shuffle implementation, e.g. other lakehouse sinks (delta lake) could have similar approach.
+
+### Caveats
+
+When iceberg partition spec only contains range partitions(e.g. year, month, day), we don't need to do this shuffle, otherwise all traffic will go to same sink.
```
fuyufjh, line 67: How to identify out these columns? so that you can exclude these columns from the user-specified partition key.

rfcs/0077-iceberg-sink-shuffle.md
```diff
@@ -0,0 +1,72 @@
...
+
+![image](images/0077-iceberg-sink-shuffle/after_shuffle.svg)
+
+This way we can reduce the number of data files in the table, which helps to improve the read performance of iceberg.
+
+### Implementation
+
+There are two possible implementations for this feature:
+
+1. Add a new `IcebergPartitionOperator`, which calculates the partition value of each record and adds it to the record, then asking the dispatcher executor to do hash shuffle according to the partition value. The plan is like following:
```
fuyufjh, line 42: IIRC, our common sinks also have such a `Shuffle` when the sink's PK doesn't match its stream key. cc. @st1page 


This is very similar with this design, the only problem is that previously we didn't distinguish **partition key** from primary key.

fuyufjh: Also cc. @xiangjinwu @tabVersion

Previously in channel `#wg-new-source-ddl` and https://github.com/risingwavelabs/risingwave/issues/9443, we decided to use the sink property `primary_key` for both PK and partition key, depending on different connectors, for example,
- Kafka upsert sink will use `primary_key` as partition key & PK
- Kafka append-only sink will use `primary_key` as partition key only

While **this RFC proposes to introduce 2 different properties: partition_key and primary_key respectively**, right?

st1page: > IIRC, our common sinks also have such a Shuffle when the sink's PK doesn't match its stream key

clarify the condition can lead to the issue: when the partition key does not contain the whole stream key.

rfcs/images/0077-iceberg-sink-shuffle/before_shuffle.svg
CAJan93, line 1: Should the partitions not have different years? In this example we partition by year, right?

liurenjie1024: They should have different years, I'm just simplify it😅

rfcs/0077-iceberg-sink-shuffle.md
liurenjie1024: I think we don't need to identify out these columns? We just need to check the partition spec?

rfcs/0077-iceberg-sink-shuffle.md
liurenjie1024: > While this RFC proposes to introduce 2 different properties: partition_key and primary_key respectively, right?

In fact, by `primary_key` here I mean `stream_key` of sink's input. I think for iceberg we don't need user to specify `primary_key` here?

st1page: The optimizer must know the downstream's partition requirements. in the past, we were just concerned about correctness issues such as "make sure the order of operations for the specific key" or "no multiple parallelisms modifying the same key, which can bring dead-lock". And in this RFC, if the downstream system has a stronger partition key to achieve better performance, we can use it as the distribution stragey, and the corectness issues should be guaranteed by the downstream system.  

About if the user needs to specify the primary_key/partition key in the `CREATE SINK` statement. 
- for kafka/redis, the primary key/partition key has not a table level's catalog or config, user must give it in `CREATE SINK` statement to define the behavior of the RW sink.
- for MySQL/JDBC, current behavior is that requiring users to specify the primary_key in the `CREATE SINK` statement. It will be validated in the meta node if the user-specified pk is the same with the JDBC catalog. It is because currently frontend can not query the JDBC's catalog without java runtime c.c. @StrikeW correct me if I am wrong
- and for iceberg in this RFC, I think user do not need to specify any information in the `CREATE SINK` because our fe node can get the information from the down stream system's catalog

rfcs/0077-iceberg-sink-shuffle.md
fuyufjh: I mean, for example, the partition key is `id, day, category`, so you need to exclude the `day` column and keep the `id, category` columns as the partition key for RW's Iceberg Sink executor, right?

rfcs/0077-iceberg-sink-shuffle.md
```diff
@@ -0,0 +1,197 @@
...
+
+```mermaid 
+flowchart TD
+    A[Source] --> B[DispatcherExecutor]
+    B -->|IcebergDispatcher| C[IcebergSink]
+```
+
+In this approach we need to add an extra `IcebergDispatcher` to dispatcher executor. The `IcebergDispatcher` will calculate the partition value of each record and do hash shuffle according to the partition value.
+
+I prefer approach 1 since it's more extensible and does not change too much current shuffle implementation, e.g. other lakehouse sinks (delta lake) could have similar approach.
```
fuyufjh, line 193: Hmm, actually I would prefer approach 2 because we consider the `Merge` and `Dispatch` as very special operators and some implementations depend on this assumption. 

------

Another thing worth to mention is: As RisingWave is the writter of downstream Iceberg, we can choose to support only a **subset** of Iceberg's partition feature. Generally, Iceberg supports partition *expressions* rather than just a list of columns. However, we may only support a list of columns as partition key, which can simplify our design.

This is also [what Flink did](https://iceberg.apache.org/docs/latest/flink-ddl/):

> PARTITION BY (column1, column2, ...) to configure partitioning, Flink does not yet support hidden partitioning.

rfcs/0077-iceberg-sink-shuffle.md
liurenjie1024: No, we don't.

rfcs/0077-iceberg-sink-shuffle.md
liurenjie1024: > Hmm, actually I would prefer approach 2 because we consider the Merge and Dispatch as very special operators and some implementations depend on this assumption.

I still prefer approach 1 for several reasons:
1. `Dispatch` executor in fact calls `Dispatcher` for different  distributions in risingwave, and I don't want to introduce another distribution which serves only external system.
2. It wastes computed partition values.

liurenjie1024: > Another thing worth to mention is: As RisingWave is the writter of downstream Iceberg, we can choose to support only a subset of Iceberg's partition feature. Generally, Iceberg supports partition expressions rather than just a list of columns. However, we may only support a list of columns as partition key, which can simplify our design.

You misunderstood flink's iceberg implementation. The fact is that it doens't support ddl statement, but the writer still respect partition spec created by other engines such as spark, you can try it yourself. 

Also I really don't like the statement that we should not do sth because  flink doesn't support it. Why users should choose use if we blindly follow flink?

ZENOTME:
There is also probability that cardinality of range only partition be ten thousands. E.g. the partition is `year, month,day,hour`.🤣 But the probability of this may be not high and we can solve it in  https://github.com/risingwavelabs/risingwave/issues/12016.

liurenjie1024:
> There is also probability that cardinality of range only partition be ten thousands. E.g. the partition is `year, month,day,hour`.🤣 But the probability of this may be not high and we can solve it in [risingwavelabs/risingwave#12016](https://github.com/risingwavelabs/risingwave/issues/12016).

Yes, that's possible in an extreme case: when we are catching up with history data.

rfcs/0077-iceberg-sink-shuffle.md
fuyufjh: > Also I really don't like the statement that we should not do sth because flink doesn't support it

No, this is not my motivation. My whole point is:

> As RisingWave is the writter of downstream Iceberg, we can choose to support only a subset of Iceberg's partition feature.

So I'd like to leverage this to simplify our design.

fuyufjh: I have some misunderstanding. Now it LGTM

PR #78, opened 2023-11-21
RFC: Decouple iceberg sink commit from risingwave checkpoint

liurenjie1024:
[Rendered](https://github.com/risingwavelabs/rfcs/pull/78/files?short_path=247cae2#diff-247cae2975e594e74c6427cb05d2fe1ae43370af3926c44ec43ebdfcfce0dbfd)


rfcs/0078-iceberg-sink-decouple-checkpoint.md
```diff
@@ -0,0 +1,127 @@
...
+
+1. This will increase failure recovery time for iceberg sink. For example, when the commit interval is set to 30 minutes, and sink failed in the 29 minute, we will need to replay all data for the first 29 minutes.
+
+#### Approach 2
+
+In this approach we don't do any modification to sink log writer, but modifies iceberg sink. Instead of committing all data files to iceberg table in every checkpoint, we flush data into parquet files, and save the flushed file paths into state table. Following graph illustrates the case:
+
+![Write parquet every checkpoint](images/0078-iceberg-sink-decouple-checkpoint/write_parquet_per_cp.svg)
+
+While this method is simple enough, we may still experience small file problems if we do checkpoint frequently. We can further decouple flushing parquet file with checkpoint. Instead of flusing parquet files in every checkpoint, we flush parquet row groups. Following diagram illustraes the process:
```
hzxa21, line 105: This can work only if the row group boundary can be completely controlled by ourselves. In other words, we can close a row group immediately when iceberge sink receives the checkpoint barrier. Is this already supported in icelake?

rfcs/0078-iceberg-sink-decouple-checkpoint.md
```diff
@@ -0,0 +1,127 @@
...
+
+![Write row group every checkpoint](images/0078-iceberg-sink-decouple-checkpoint/write_row_group_per_cp.svg)
+
+There are chances we don't even need to flush row group. For example we can save the record sequence id of current row in log store to skip flusing row group, but I don't introduce to much dependency on such characteristics of log store to make things more complicated. One row group only adds a record in parquet's `FileMetaData`, and it has no impact on other readers of parquet.
```
hzxa21, line 109: > save the flushed file paths into state table.

I think we can record the current write position in the state table instead of log store.

ZENOTME:
> In this approach we don't do any modification to sink log writer, but modifies iceberg sink. Instead of committing all data files to iceberg table in every checkpoint, we flush data into parquet files, and save the flushed file paths into state table. Following graph illustrates the case:

In this case, maybe we don't need to flush data into parquet files every checkpoint to avoid small file.🤔

rfcs/0078-iceberg-sink-decouple-checkpoint.md
```diff
@@ -0,0 +1,127 @@
...
+Pros of this approach:
+
+1. Easier to implement.
+2. Easier to apply to more sinks (such as delta lake).
+
+Cons of this approach:
+
+1. This will increase failure recovery time for iceberg sink. For example, when the commit interval is set to 30 minutes, and sink failed in the 29 minute, we will need to replay all data for the first 29 minutes.
+
+#### Approach 2
```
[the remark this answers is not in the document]
liurenjie1024: It's a little more complicated, but not that much. We can further discuss this later when we want to take approach 2.

liurenjie1024:
Conclusion: we will take approach 1.

fuyufjh (approved):

PR #79, opened 2023-11-22
RFC: use `include ... as ...` to ingest more message parts

tabVersion:


rfcs/0079-include-key-as.md
```diff
@@ -0,0 +1,89 @@
...
+    [ include ... [ as ... ] ]
+with ( ... )
+format ... encode ... ( ... )
+[ key encode [ ( ... ) ] ]
+```
+
+* If `as` name is not specified, a connector-component naming template will be applied
+  * For connector kafka and component key, the derived message key column name is `_rw_kafka_key`.
+* The default type for message key column is `bytea`. The priority of the type definition is: 
+  `key encode` > infer from `format ... encode ...` > default type
```
st1page, line 51: > infer from `format ... encode ...`

What dose it mean?

tabVersion: if there is a schema registry and we can infer the key schema from it.

rfcs/0079-include-key-as.md
```diff
@@ -0,0 +1,89 @@
...
+[ key encode [ ( ... ) ] ]
+```
+
+* If `as` name is not specified, a connector-component naming template will be applied
+  * For connector kafka and component key, the derived message key column name is `_rw_kafka_key`.
+* The default type for message key column is `bytea`. The priority of the type definition is: 
+  `key encode` > infer from `format ... encode ...` > default type  
+* **Important**: `include key` is required for `format upsert` and RisingWave will use the key column as one and 
+  only primary key to perform upsert semantic. It does not allow to specify multiple columns as primary key
+  even if they are part of the key.
```
fuyufjh, line 54: So,

- if I specify `key encode avro` and `include key as x`, `x` will be `struct`
- if I specify `key encode json` and `include key as x`, `x` will be `JSONB`

Right?

rfcs/0079-include-key-as.md
```diff
@@ -0,0 +1,89 @@
...
+
+In previous implementation, we only load message key into a [~~hidden~~](https://github.com/risingwavelabs/risingwave/pull/13521) 
+column called `_rw_key` as `bytea` on default.
+
+Here are some cases when RisingWave derives the `_rw_key` column.
+
+* For all connectors with `format upsert`, RisingWave derives the column as primary key to perform upsert semantic.
+  * An exception: for `format upsert encode json`, RisingWave allows the PK to be part of the message payload instead of
+  the message key. This behavior is more like `format plain encode json` with PK constraint.
+  * **To Be Discussed**: whether we can afford to perform a breaking change for the above exception.
```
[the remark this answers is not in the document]
fuyufjh: > * An exception: for `format upsert encode json`, RisingWave allows the PK to be part of the message payload instead of the message key. This behavior is more like `format plain encode json` with PK constraint.

IIRC, according to our previous discussion, this will be `format insert` + normal PK definition i.e. `primary key (foo, bar)`. 

The exception looks very inconsistent and I'd like to get rid of it.

fuyufjh (commented):
Generally LGTM

rfcs/0079-include-key-as.md
```diff
@@ -0,0 +1,89 @@
...
+| offset             | `i64`                                      | The offset in the partition                       |
+| header             | `struct<varchar, bytea>[]`                 | KV pairs along with message                       |
+
+### Pulsar
+
+| Allowed Components | Default Type | Note                                                                                      |
+|--------------------|--------------|-------------------------------------------------------------------------------------------|
+| key                | `bytea`      | Allow overwritten by `encode` and `key encode`. Refer to `MessageMetadata::partition_key` |
+
+More components are available at [here](https://docs.rs/pulsar/latest/pulsar/message/proto/struct.MessageMetadata.html).
```
xiangjinwu, line 74: How do we decide which metadata fields to expose? Do we intend to expose similar (not sure if same) concepts from different connectors using the same word? For example
* kafka `key`, pulsar `partition_key`, kinesis `partition_key`
  * There is also pulsar `ordering_key` or kinesis [ExplicitHashKey](https://docs.aws.amazon.com/kinesis/latest/APIReference/API_PutRecord.html#API_PutRecord_RequestSyntax)
* kafka `offset`, pulsar `sequence_id`, kinesis `sequence_number`
  * i64 vs u64 vs string
* kafka `timestamp`, pulsar `publish_time`, kinesis `approximate_arrival_timestamp`
  * There is also pulsar `event_time`
* kafka `headers`, pulsar `properties`
  * kafka headers is `struct<varchar, bytea>[]` but pulsar properties is `struct<varchar, varchar>[]`

rfcs/0079-include-key-as.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Kafka
+
+| Allowed Components | Default Type                               | Note                                              |
+|--------------------|--------------------------------------------|---------------------------------------------------|
+| key                | `bytea`                                    | Allow overwritten by `encode` and `key encode`    |
+| timestamp          | `timestamp with time zone` (i64 in millis) | Refer to `CreateTime` rather than `LogAppendTime` |
+| partition          | `i64`                                      | The message is from which partition               |
+| offset             | `i64`                                      | The offset in the partition                       |
+| header             | `struct<varchar, bytea>[]`                 | KV pairs along with message                       |
```
xiangjinwu, line 66: plural: `headers`

rfcs/0079-include-key-as.md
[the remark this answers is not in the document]
st1page: In some cases, user want to determine the primary key of the table for some use cases such as 
- better batching/ serving performance
- streaming temporal join
And in those cases, user can make sure the primary keys column is exactly the message key in Kafka.

xiangjinwu:
Would like to also mention an alternative syntax:
```
create source/table (
    ..,
    <key-column> bytea FROM SOURCE key,
    <timestamp-column> timestamptz FROM SOURCE timestamp,
    <headers-column> struct<key varchar, value bytea>[] FROM SOURCE headers,
    primary key ( <key-column> )
)
with ( ... )
```

Pros and cons of this syntax, compared to `include ... as ...`:
* All columns are listed in one place, similar to DEFAULT and GENERATED columns.
* Users are required to spell out the data type.
* (welcome to add more)

rfcs/0079-include-key-as.md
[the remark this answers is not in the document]
hzxa21: If `format upsert` is specific to MQ with delete tombsonte (null value), I think the PK must be the MQ message key and cannot be a field in the value.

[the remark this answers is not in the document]
fuyufjh: I think either of the following behavior is acceptable:

1. All fields in message key are used as PK, and empty message body will be considered as delete tomb 
2. Some fields in message body are used as PK, and empty message body will not be considered as delete tomb.

1 corresponds to `upsert` semantics, while 2 corresponds to `plain`/`insert` semantics

rfcs/0079-include-key-as.md
fuyufjh: > Do we intend to expose similar (not sure if same) concepts from different connectors using the same word?

I think it's not necessary to unify them. If there is ambiguity, I tend to let them have different names.

st1page:
> Would also like to mention an alternative syntax:
> 
> ```
> create source/table (
>     ..,
>     <key-column> bytea FROM SOURCE key,
>     <timestamp-column> timestamptz FROM SOURCE timestamp,
>     <headers-column> struct<key varchar, value bytea>[] FROM SOURCE headers,
>     primary key ( <key-column> )
> )
> with ( ... )
> ```
> 
> Pros and cons of this syntax, compared to `include ... as ...`:
> 
> * All columns are listed in one place, similar to DEFAULT and GENERATED columns.
> * Users are required to spell out the data type.
> * (welcome to add more)

Currently, we do not allow users to define schema in column clauses when using schema registry or schema file 
Another method is using the "star" grammer
https://github.com/risingwavelabs/risingwave/issues/12209

rfcs/0079-include-key-as.md
```diff
@@ -0,0 +1,89 @@
...
+
+```sql
+create source/table (
+    ..,
+    primary key ( <key-column> )
+)
+    [ include key [ as <key-column> ] ]
+    [ include timestamp [ as <timestamp-column>] ]
+    [ include header [as <header-column>] ]
+    [ include ... [ as ... ] ]
```
BugenZhao, line 42: Have you considered the generated column syntax for this? For example, calling dummy functions to get the meta data.

fuyufjh: Generated column asks users to specify the data type, which might be a little bit challenging sometimes, such as https://github.com/risingwavelabs/rfcs/pull/79#discussion_r1402891627

fuyufjh:
FYI. Feel free to join the slack channel `#wg-include-key-as`

rfcs/0079-include-key-as.md
xxchan: Update: now we do want to unify `partition` and `offset` for them.. @tabVersion Can you add the rationale for that in the RFC?  😇

rfcs/0079-include-key-as.md
```diff
@@ -0,0 +1,94 @@
...
+More components are available at [here](https://docs.rs/pulsar/latest/pulsar/message/proto/struct.MessageMetadata.html).
+
+### Kinesis
+
+| Allowed Components | Default Type                             | Note                                                                             |
+|--------------------|------------------------------------------|----------------------------------------------------------------------------------|
+| key                | `bytea`                                  | Allow overwritten by `encode` and `key encode`. Refer to `Record::partition_key` |
+| timestamp          | `timestamp with time zone` (from chrono) | refer to `Record::approximate_arrival_timestamp`                                 |
+| partition          | `varchar`                                    | The message is from which partition                                              |
+| offset             | `varchar`                                    | The offset in the partition                                                      |
```
xxchan, line 85:
```suggestion
| offset             | `varchar`                                    | The offset in the partition. Corresponds to Kinesis sequence numbers.                                                     |
```

Users might feel confused about the concept's name. At least we might need to document it clearly.

xxchan: BTW, I'm thinking that whether the default name can be different, even if we want the unified data type.

I think we will rely on `AdditionalColumnType` when the column is needed, instead of the col name.

rfcs/0079-include-key-as.md
```diff
@@ -0,0 +1,89 @@
...
+|--------------------|--------------|-------------------------------------------------------------------------------------------|
+| key                | `bytea`      | Allow overwritten by `encode` and `key encode`. Refer to `MessageMetadata::partition_key` |
+
+More components are available at [here](https://docs.rs/pulsar/latest/pulsar/message/proto/struct.MessageMetadata.html).
+
+### Kinesis
+
+| Allowed Components | Default Type                             | Note                                                                             |
+|--------------------|------------------------------------------|----------------------------------------------------------------------------------|
+| key                | `bytea`                                  | Allow overwritten by `encode` and `key encode`. Refer to `Record::partition_key` |
```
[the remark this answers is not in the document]
tabVersion: Yes, I make it `bytea

rfcs/0079-include-key-as.md
tabVersion: > BTW, I'm thinking that whether the default name can be different, even if we want the unified data type.

of course, the `offset`/`partition` here serves as an identifier where Risingwave loads the field. The default name should be `_rw_{connector name}_offset`.
And there is an `as` clause following the identifier, allowing renaming the column. 

> I think we will rely on AdditionalColumnType when the column is needed, instead of the col name.

Yes, it exactly what I design

rfcs/0079-include-key-as.md
tabVersion: Yes, we are going to unify the partition and offset column type for all connectors.
The basic reason is that we want to impl a source exec level throttling and requires a chunk to be cut anywhere with maintaining the offset info. Besides, the source backfill feature also rely on the behavior to tell when to end the backfill stage.

st1page (approved):

PR #81, opened 2023-12-07
Parameters for each streaming job

st1page:


TennyZhuang:
First of all, we even don't have a Per mview config. The RFC seems good to me but lack of implementation detail.

st1page:
> First of all, we even don't have a Per mview config. The RFC seems good to me but lack of implementation detail.

which part :eyes: I even tell you where we should store the config proto

st1page:
> I'm also working on an RFC, which try to introduces dynamic global config :(

I think there is no conflict because this RFC just focuses on the runtime parameter in the executor and do not care about other components' config for a Compute node or the system parameters mentioned in this RFC https://github.com/risingwavelabs/rfcs/pull/34

rfcs/0081-dynmical-config-for-stream-job.md
```diff
@@ -0,0 +1,81 @@
...
+3. alter in each executor: every stream executor must can receive the alter config barrier and adopt the new config.
+4. The issue of the barrier based alter is that the barrier latency could be large and in that case, the config can not be altered in timely manner. https://github.com/risingwavelabs/risingwave/issues/13803 
+   1. So we can introduce a control channel to all the executors like this https://github.com/risingwavelabs/risingwave/pull/4834 so that the executor can get the 
+   2. another method is store a config in the task local and executor should check if the config in the task local has been changed periodically(maybe with a config version). 
+
+The second is too complex to implement and can also introduce performance performance degradation. The 4.1 control channel will intrude every executor's logic. So personally, I'd like to eliminate them first.
+
+And the 3,4 maybe can not satisfy all the config and the executor can only change the config with recreating itself.
+
+I think we can implement the recovery based alter first and optimize the configs with 3 or 4.2 later.
```
kwannoel, line 81: +1 for recovery based alter first, and optimize with 4.

3 is easily implemented (and is currently implemented for rate limit), but it can cause delays in config change if barrier latency is high. So -1 for it.

rfcs/0081-dynmical-config-for-stream-job.md
```diff
@@ -0,0 +1,81 @@
+---
+feature: Dynamic Config For Stream Job
+authors:
+  - "st1page"
+start_date: "2023/12/07"
+---
+
+# Dynamic Config For Stream Job
+
+Thanks all the people took part in the discussion of https://github.com/risingwavelabs/risingwave/issues/11929 and implantation of `stream_rate_limit`. The RFC is hight inspired by them.
```
kwannoel, line 10:
```suggestion
Thanks all the people took part in the discussion of https://github.com/risingwavelabs/risingwave/issues/11929 and implementation of `stream_rate_limit`. The RFC is inspired by them.
```

rfcs/0081-dynmical-config-for-stream-job.md
```diff
@@ -0,0 +1,81 @@
...
+In other database, those parameter is declared by session variable or session config. It is because the compute task is bounded in those batch system and it is not necessary to alter the config for a running job. But for a long-running streaming job, alter the config for a created and processing job is needed.
+
+This RFC will introduce a dynamic config framework to make user
+- can declare those configuration when create streaming job
+- can alter those configuration for a long-running streaming job. 
+### Non-Goals
+
+The RFC will not concern those config can change the streaming graph and its topology, such as
+- The configs that can change the plan such as `RW_TWO_PHASE_AGG`
+- The configs that can change the actor's number such as `STREAMING_PARALLELISM` (maybe we can use the same syntax to alter it later, but it will be only a semantic sugar of scale)
```
kwannoel, line 38:
```suggestion
- The configs that can change the actor's number such as `STREAMING_PARALLELISM` (maybe we can use the same syntax to alter it later, but it will be only a syntactic sugar of scale)
```

kwannoel (approved):
LGTM, the recovery based alter is a low-hanging fruit we can support first.

rfcs/0081-dynmical-config-for-stream-job.md
```diff
@@ -0,0 +1,81 @@
...
+  - "st1page"
+start_date: "2023/12/07"
+---
+
+# Dynamic Config For Stream Job
+
+Thanks all the people took part in the discussion of https://github.com/risingwavelabs/risingwave/issues/11929 and implementation of `stream_rate_limit`. The RFC is inspired by them.
+
+## Background
+Currently, we have lots of config, switch and parameter for the streaming execution. They often used for performance optimization for different workload. But they distributed different places currently.
```
xxchan, line 13: What part of the following will be unified (replaced) with the new mechanism?

st1page: I will rewrite the part later

st1page: I think all the cases in the example are able to be implemented as the new mechanism, but we can implement them on-demand

xxchan: For constant value I think that's easy and we can delay. 

For others that are already configurable by some mechanisms now, I'm mainly interested in migration/deprecation plan. e.g., for `OverWindowCachePolicy`, on one hand it's persisted in the plan, and on the other hand it's controlled by a session variable (but not documented yet? so less likely a large concern). 

I think unify them with the new mechanism can reduce users' mental burden. Just want to see these discussed.

st1page: Good question! I will think and add the methods to migration for these different cases.

rfcs/0081-dynmical-config-for-stream-job.md
```diff
@@ -0,0 +1,81 @@
...
+
+First to clarify, there is some parameter can not be altered such as `timezone`. Because it can change the result of the streaming query in SQL semantic. It is strange to alter such a config for a "running query", it is a little like alter mv which change the SQL statement.
+
+User will be able to alter the dynamic stream config for specified MV or SINK with the grammar.
+```SQL
+ALTER [MATERIALIZED VIEW | SINK] SET parameter TO [value | 'value'];
+```
+The statement will send a RPC to meta. After reducing the config in the common place, it is easier to alter the config of some relations. And we need some implementation complexity trade-off here.
+
+1. alter with recovery: change the configs in the meta store and trigger a recovery. Then the newly created actor will adopt the new configs.
```
tabVersion, line 70: changing config seems too heavy to apply. Though it is the easiest way, it is counter-intuitive that users change sth and get lost from the cluster, even for several seconds. 
I don't get the position for the proposal if it is an SQL command, I don't think it can alter the cluster's status.
If it is a DevOps tool, it should not be released as a command. I think meta command is a better way.

kwannoel: >  get lost from the cluster, even for several seconds.
I don't think we need to trigger the recovery, we can let user manually trigger it.

The UX will be:
1. user alters the MV: `alter materialized view mv1 with ( ... )` (perhaps need a section on the exact syntax)
2. the user receives a response + a notice: `NOTICE: Please restart the cluster with TRIGGER RECOVERY / risectl meta recovery for your changes to take effect`
3. The user runs the recovery command.

So it would be explicit and opt-in by the user

st1page: > I don't get the position for the proposal if it is an SQL command.

So how about the 4th method :thinking: do you think it is acceptable if we can change the log without recovery?

tabVersion: yes, I think 4 is the ideal way for such change. I think there are two phases: 1) change the proto in etcd and  commit to meta, it is safe to restart the cluster, 2) tell the user that the change has been applied. 

> the user receives a response + a notice: NOTICE: Please restart the cluster with TRIGGER RECOVERY / risectl meta recovery for your changes to take effect

If so, we are encouraging the users to test with risectl, a not well-documented tool designed for advanced users and the dev team. I mean doing anything in SQL, then it should be simple and concise, without introducing extra tools or doing the devops' job.

st1page: > > the user receives a response + a notice: NOTICE: Please restart the cluster with TRIGGER RECOVERY / risectl meta recovery for your changes to take effect
> 
> If so, we encourage users to test with risectl, a poorly well-documented tool designed for advanced users and the dev team. I mean doing anything in SQL, should be simple and concise, without introducing extra tools or doing the devops' job.

I guess it is more like the `ALTER SYSTEM` statement
>Values set with ALTER SYSTEM will be effective after the next server configuration reload, or after the next server restart in the case of parameters that can only be changed at server start. A server configuration reload can be commanded by calling the SQL function pg_reload_conf(), running pg_ctl reload, or sending a SIGHUP signal to the main server process.

tabVersion:
A crazy idea for this 😈
Dump a detailed yml for the compute graph (at the level of executors) and let the users modify the yml and then apply them back. A per streaming job config is too complex, for example, a hashagg and a table scan have diff configs. Covering all configs with CLI can cost a lot of work.
So let users try on the risk of themselves seems a smarter way. Then we can build some tools to make the process safer.

BugenZhao:
Personally I doubt if there are enough benefits to justify doing this. 😕

To be honest, configurations like `extreme_cache_size` or `hash_agg_max_dirty_groups_heap_size` are introduced just because we have no idea about the optimal values they should have. Making them configurable does give some kind of flexibility to users but I don't think most of them will ever touch them. If we really want to be serious now and expect some rare but advanced users appearing and willing to customize them, the granularity of streaming job can be unfortunately **far from sufficient**.

For example, why all `HashAgg` in the streaming job must have the same value of `extreme_cache_size` that is specified in `WITH` options? The target users have been so advanced to tweak such magical numbers for peak performance, why would they mind taking control of everything? To me, this can be somehow compared to the optimizer hints from modern RDBMS, where users will specify the algorithm or execution plan they would like to have for each `JOIN` or aggregation functions in the SQL query. Advanced users may write something like the following, which is unnecessarily complicated and overkilling. However, I don't believe the compromise in this RFC truly makes sense in this case.

```sql
CREATE MATERIALIZED VIEW mv AS SELECT
/*+ max_dirty_groups_heap_size(114514) */
MAX(a) /*+ extreme_cache_size(10) */,
MIN(b) /*+ extreme_cache_size(50) */
FROM ..
```

On the other side, there are already some configurations that can be tweaked in a fine granularity like actors or executors in our system, e.g. `rate_limit`. One may find the previous arguments not convincing, but making rate limit a job-level configuration definitely leads to a regression that users will not be able to set different rate limits for different sources anymore. IMO, the best places they should reside are simply the plan nodes they belong to. One may argue that we can introduce some dynamic into the job-level configuration like `rate_limit.source.1001 = 1000`, but what about `OverWindowCachePolicy` (or again, `extreme_cache_size`)? How could users find a way to specify which aggregation call they're referring to? It appears that we have to adopt the Hint approach again.

In short, I find the design working hard to unify the configuration entries into a way that does not fit the best for every kind of them. I think we should focus on the semantics and functionalities first before talking about the implementation like how to apply the changes online.

st1page:
> It appears that we have to adopt the Hint approach again.

I think the hint is not a good way for the long-running streaming jobs because modify the changes online seems a required feature for us.

> In short, I find the design working hard to unify the configuration entries into a way that does not fit the best for every kind of them. I think we should focus on the semantics and functionalities first before talking about the implementation like how to apply the changes online.

I prefer to making progress little by little. At least an streaming-job-level config is much better than a constant value or global config. And I think we can build the framework first and see which config is suitable for the framework one by one.

For the more fine-grained config, we can do it with the framework too. For example.

```
with {
rate_limit.source.1001 = 1000
agg.extreme_cache_size.{executor_id}.{agg_call_idx}
}
```

st1page:
For the more fine-grained config. How about use Hint to config it when creating stream job but change it after the streaming job having created? Because for the existent MV, we have id to identify the more fine-grained objects

ALTER MV mv ADD with config
with {
rate_limit.source.1001 = 1000
agg.extreme_cache_size.{executor_id}.{agg_call_idx}
}

st1page:
> For the more fine-grained config. How about use Hint to config it when creating stream job but change it after the streaming job having created? Because for the existent MV, we have id to identify the more fine-grained objects
> 
> ```
> ALTER MV mv ADD with config
> with {
> rate_limit.source.1001 = 1000
> agg.extreme_cache_size.{executor_id}.{agg_call_idx}
> }
> ```

Yes, you are right. @BugenZhao  At least RW does not expose some terminology to users and does not have the system table for is.
Such as "operator" which means a set of executors wihch is the same node on different parallelisms.

st1page:
After some offline discussion, we all think there are lots of issues of the RFC. And because this PR is only a draft that is far from complete. @stdrc will give draft new RFC later.

BugenZhao:
> > For the more fine-grained config. How about use Hint to config it when creating stream job but change it after the streaming job having created? Because for the existent MV, we have id to identify the more fine-grained objects
> > ```
> > ALTER MV mv ADD with config
> > with {
> > rate_limit.source.1001 = 1000
> > agg.extreme_cache_size.{executor_id}.{agg_call_idx}
> > }
> > ```
> 
> Yes, you are right. @BugenZhao At least RW does not expose some terminology to users and does not have the system table for is. Such as "operator" which means a set of executors wihch is the same node on different parallelisms.

I guess a Graphic User Interface (GUI) would be more suitable in this scenario. Imagine that users locate the desired operator in the fragment graph on the web dashboard then directly edit some fields to apply changes. 🤣

kwannoel:
> After some offline discussion, we all think there are lots of issues of the RFC. And because this PR is only a draft that is far from complete. @stdrc might give draft new RFC later.

Perhaps we can have a draft for just the control channels part.

PR #83, opened 2023-12-15
change log tvf

st1page:


st1page:
the idea is very same with "What's the Difference? Incremental Processing with Change Queries in Snowflake" https://dl.acm.org/doi/pdf/10.1145/3589776
Need to check later

st1page:
https://docs.risingwave.com/sql/commands/sql-as-changelog#as-changelog

PR #85, opened 2024-01-12
RFC: Combine Historical and Incremental Data

fuyufjh:
Migrated from Notion.

### DIFF
--- a/rfcs/0085-iceberg-source.md
+++ b/rfcs/0085-iceberg-source.md
@@ -1,0 +1,129 @@
+---
+feature: iceberg_source
+authors:
+  - "Eric Fu"
+start_date: "2024/1/5"
+---
+
+# RFC: Combine Historical and Incremental Data
+
+## Motivation
+
+Some users need to create a table that combines data from historical data (such as file source or Iceberg Source) and incremental data (such as Kafka or Pulsar)
+
+Secondly, We’d like to allow users to analyze data from Iceberg or other data stores directly. As a product strategy, we would offer more ability to run analysis on historical using RisingWave.
+
+## Design
+
+The design comprises the following parts.
+
+1. **Iceberg Source**, which is queryable but unable to be base table of streaming jobs. 
+2. **Alter the connector of a table**
+
+With these new features, users will be able to 
+
+1. Create an iceberg source for historical data
+2. Create a RisingWave table and insert data selected from the iceberg source
+3. Alter the table to add connector from the upstream Kafka
+
+With this design, we give users more flexibility. For examples,
+
+- Case 1: if the historical data and Kafka data have different schema (e.g. one in relational table and another in JSON), they would be able to restructure the data with SQL.
+- Case 2: If the historical data is from another Kafka topic, they can simply create a table with the old Kafka connector, and then alter it to another new Kafka topic.
+- Case 3: If the users want to load a snapshot from Postgres but not sync data in the future, they can create a PG-connected table and alter the connector to none after the data is loaded.
+
+### 1. **Iceberg Source**
+
+Iceberg Source introduces a new category of source - queryable but unable to be base table of streaming jobs. The reason behind this is straightforward: Apache Iceberg as well as other data lake projects does not provide a good interface for CDC, and it doesn’t sound like a good practice because data lakes are usually treated as the final stop of data. 
+
+> **Source or External Table?**
+> 
+> Shall we introduce this “queryable but not streamable” thing as external table? 
+>
+> **Pros:**
+> - It clearly shows that it’s “queryable but not streamable”
+>
+> **Cons:**
+> - It’s a new concept
+> - Kafka source is already queryable (and it turns out to be very useful), so the concept is somehow overlapped with source. So does the file source.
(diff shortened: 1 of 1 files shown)


st1page:
We encountered an interesting problem: what happens when users want to define a watermark on this table? If we directly apply the ideas from this RFC, because the order of data inserted into the table by batch queries is unordered, unexpected records will likely be expired and deleted by the watermark. I think there are a few possibilities in my thoughts.

1. (Completely unfeasible) Define a watermark only on the Kafka source and sink it into the table. This actually cannot achieve the semantics that users want. Because the table after the source will erase all the watermark semantics that users want, and the watermark cannot be used downstream.
2. Use `insert ... select ... order by time asc` to insert data into the table with a watermark definition in time order. In this case, we may need to ensure these:
    - In the shuffle from BatchInsertExecutor to StreamDMLExecutor, we need to adopt a sort merge or no shuffle method to ensure that the order of the insertion stream is the same as the output of the batch, maintaining the order.
    - (Optimization) We need to implement a sort scan from the Iceberg source (push down of order by).
3. Support the `ALTER TABLE ... ADD WATERMARK ...` syntax. In this method, users need to first import historical data into a table without a watermark. Then add a watermark to the table, and subsequently construct the upstream (subscribe to changes from Kafka) and downstream (stream processing logic). The problem with this approach is that when we have a materialized view on the table, we cannot send watermarks downstream during the backfill of historical data, which is inefficient or even unacceptable for some stream operators that rely on watermarks. If users use features related to EOWC (End of Watermark Collection), they may also have to accept sending a large amount of data downstream when the first watermark arrives.

fuyufjh:
The 3rd proposal `ALTER TABLE ... ADD WATERMARK ...` sounds very limited. It requires the table has no downstream streaming jobs associated, which asks the users to prepare everything before doing it in fact. It works but not quite friendly.

I slightly prefer the 2nd proposal `insert ... select ... [order by time asc]`. In my mind, the `order by` clause here is not enforced. For example, if the users ingested data from Kafka source first e.g. a topic named `historical_events`, the events will be almost ordered naturally. Furthermore, if Iceberg can provide such *almost ordered* reading, it's also doable. (According to [Iceberg - Flink queries](https://iceberg.apache.org/docs/1.5.1/flink-queries/#emitting-watermarks) docs, perhaps there is such a method)

xxchan (commented):
Some new ideas by @st1page  from https://risingwave-labs.slack.com/archives/C07CU2YBKCG/p1721184731055789

Since we are adding batch read _function_ https://github.com/risingwavelabs/risingwave/pull/17673, we can combine a batch query with a connector. Then we don't need a batch source and no need to `ALTER TABLE` any more.

tentative syntax:
```sql
CREATE TABLE orders (
    order_id INT,
    customer_name VARCHAR,
    data JSONB,
    PRIMARY KEY (order_id, customer_name)
) INITIAL WITH SELECT * FROM file_scan(
  'parquet',
  's3',
  'ap-southeast-2',
  'xxxxxxxxxx',
  'yyyyyyyy',
  's3://your-bucket/path/to/*'
);
WITH (
  connector = 'kinesis',
  stream = 'wkx-dynamo-orders',
  scan.startup.mode='earliest',
  aws.region = 'us-east-1',
  kinesis.credentials.access = 'ABCDEFG',
  kinesis.credentials.secret = 'abcdefg',
) FORMAT DYNAMODB_CDC ENCODE JSON;
```

@xiangjinwu : Can be achieved by `pause_on_create` + `insert into t select` + `resume`

fuyufjh:
> we can combine a batch query with a connector. Then we don't need a batch source and no need to `ALTER TABLE` any more.

Is it just `CREATE TABLE AS <select-query>`?

Here, taking your example, the columns in the table definition and the columns in `SELECT * ` actually duplicates. To eliminate the duplication, we will get something like `CREATE TABLE AS`.

xxchan:
> Is it just `CREATE TABLE AS <select-query>?`

Yes, I asked the same question. 😄 @st1page feels for the specific needs, the syntax CTAS is weird, so he wants to introduce a separated syntax.

Specifically,
- `INITIAL WITH` has an order
- We can do `INITIAL WITH` for source. The syntax CTAS is reasonable, but CSAS looks weird (to him)

fuyufjh:
> * `INITIAL WITH` has an order

Order is not that important when processing historical data. Particularly, considering multiple parallelism, the order might be less useful to users.

> * We can do `INITIAL WITH` for source. The syntax CTAS is reasonable, but CSAS looks weird (to him)

I feel `CREATE SOURCE + INITIAL WITH` brings more confusing if you take backfilling into consideration. 

-------

Hmmm, overall, I feel this is not better than the idea of `CREATE TABLE/SOURCE` with `pause_on_create = true` 😀

st1page:
> > * `INITIAL WITH` has an order
> 
> Order is not that important when processing historical data. Particularly, considering multiple parallelism, the order might be less useful to users.

> Hmmm, overall, I feel this is not better than the idea of `CREATE TABLE/SOURCE` with `pause_on_create = true` 😀

LGTM. detailly we need 
- a with option `pause_on_create`
- a new statement `RESUME table_name;`

And we might need to discuss about insert statment on the source later.

PR #88, opened 2024-05-09
RFC: Support SELECT on SUBSCRIPTION

hzxa21:
Extend the subscription concept to support:
- Batch select on subscription to inspect the snapshot of the changelog of a MV/Table
- Streaming query on subscription to facilitate alerting and changelog sink use cases.

[Preview](https://github.com/risingwavelabs/rfcs/blob/33668391026e050ee9232a159d190c0307b2f6ef/rfcs/00xx-select-from-subscription.md)

### DIFF
--- a/rfcs/00xx-select-from-subscription.md
+++ b/rfcs/00xx-select-from-subscription.md
@@ -1,0 +1,188 @@
+---
+feature: select_from_subscription
+authors:
+  - "hzxa21"
+start_date: "2024/05/08"
+---
+
+# 
+
+## Motivation
+
+This RFC aims to provide a solution to enable user to run query on a changelog stream from a MV/Table, like https://github.com/risingwavelabs/rfcs/pull/83,
+but with a different syntax. It uses the `subscription` concept instead of a TVF to represent the changelog stream.
+
+Let's first recap on the current usage on `subscription`:
+```SQL
+CREATE TABLE t (pk int PRIMARY KEY, v int);
+
+CREATE SUBSCRIPTION sub FROM t WITH (
+    retention = '1D'
+);
+
+DECLARE cur SUBSCRIPTION CURSOR FOR sub;
+
+FETCH NEXT FROM cur;
+  pk   |   v   | op | rw_timestamp 
+-------+-------+----+--------------
+ 25184 | 25184 |  1 |             
+(1 row)
+
+FETCH NEXT FROM cur;
+ t.pk  |  t.v  | t.op | rw_timestamp  
+-------+-------+------+---------------
+ 18704 | 18704 |    2 | 1715021447040
+(1 row)
+
+...
+
+```
+
+Basically we support `FETCH NEXT FROM <cursor>` on a `subscription` to get the changelog stream with schema `(original schema | op | rw_timestamp)`.
+
+
+This RFC will extend the usage of `subscription` to support `SELECT` query on it as well. The output schema of the SELECT query is `(original schema | op | rw_timestamp)`.
+
+## Design
+### Batch SELECT on subscription
+```SQL
+CREATE TABLE t (pk int PRIMARY KEY, v int);
+
+INSERT INTO t values (1, 1);
+
+CREATE SUBSCRIPTION sub FROM t WITH (
+    retention = '1D'
+);
+
+
+INSERT INTO t VALUES (2, 2);
+
+UPDATE t SET v = 11 WHERE pk = 1;
+
+# op = 1 insert
+# op = 2 delete
+# op = 3 update_insert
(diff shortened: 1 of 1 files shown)


rfcs/00xx-select-from-subscription.md
```diff
@@ -0,0 +1,188 @@
...
+SELECT * FROM sub;
+
+CREATE SINK t_changelog_sink
+AS SELECT t.pk as pk, t.v as v, CASE WHEN t.op = 2 or t.op = 4 THEN TRUE ELSE FALSE END as is_delete FROM sub;
+```
+
+- User can create sink/mv on the subscription to get the changelog stream of the MV/Table the subscription is created upon.
+- `retention` is optional for subscription eligble for streaming query.
+  - When `retention` is set, the corresponding log store will be created and the subscription is eligble for cursor query, batch select query, and streaming query.
+  - When `retention` is not set, no log store will be created for the subscription and the subscription is only eligble for streaming query.
```
hzxa21, line 114: Discussion: should we have a default retention like 10mins when retention is not set for convenience to make all subscriptions eligible for cursor query, batch select query and streaming query?

rfcs/00xx-select-from-subscription.md
```diff
@@ -0,0 +1,188 @@
...
+```
+
+Use case: one-shot inspection on the MV/Table changelog for debugging
+
+
+### Streaming SELECT on subscription
+```SQL
+CREATE TABLE t (pk int PRIMARY KEY, v int);
+
+CREATE SUBSCRIPTION sub FROM t;
```
wenym1, line 102: Will we persist the created subscription to catalog?

It seems wired to persist it to catalog. Users were intended to only create the sink, and the subscription created here is only a sugar to express the semantic of change log, which is more like a temporary thing. If we persist it, users will have to first inspect whether there are some created subscription, and then create the subscription, and then create the sink, and if not, users can just create the subscription without any concern, and then create the sink.

If we won't persist it, maybe we can make the subscription created as a per-session temporary object, or express subscription as CTE, like `WITH sub AS SUBSCRIPTION from t create materialized view as select * from sub`

rfcs/00xx-select-from-subscription.md
```diff
@@ -0,0 +1,188 @@
...
+
+  CREATE SUBSCRIPTION windowed_result_sub from windowed_result;
+
+  # op = 1: insert
+  CREATE MATERIALIZED VIEW error_minutes as 
+  SELECT 
+    distinct on window_start,
+    window_start,
+  FROM windowed_result_sub 
+  WHERE error_count > all_count * 0.9 AND op = 1;
```
neverchanje, line 149: Off-topic: maybe we can introduce a static variable, e.g, `CHANGE_INSERT` with the value 1 to avoid using magic constants everywhere.

rfcs/00xx-select-from-subscription.md
neverchanje: In my initial design, `retention` is a mandatory parameter to ensure that users are aware of the data cleaning.
I still believe that it's a right design choice.

neverchanje (approved):
Cannot find a fault in this proposal.
Overall, I think it widens the scope of our subscription ability.
I just want to explore more use cases that can utilize this feature.

rfcs/00xx-select-from-subscription.md
hzxa21: For the "create sink/mv from subscription" use case, we don't necessarily need to buffer the changelog data under global checkpoint. The reason why I make `retention` optional in this use case is because we can skip creating the log store for subscription and in this case there is no data to clean.

rfcs/00xx-select-from-subscription.md
neverchanje: I think we have already designed subscription semantically as a persistent object. So for now, it must be visible accross multiple sessions. 

But I agree with you that we can specifically introduce a per-session, "temporary subscription" for this user case to address the inconvenience, like:

```

CREATE TABLE t (pk int PRIMARY KEY, v int);

CREATE TEMPORARY SUBSCRIPTION sub FROM t;


CREATE SINK t_changelog_sink
AS SELECT t.pk as pk, t.v as v, CASE WHEN t.op = 2 or t.op = 4 THEN TRUE ELSE FALSE END as is_delete FROM sub;
```

xxhZs: If I understand correctly, we should need an execute to convert the op column of streamchunk into a changelog, so persistence is only a syntactic difference (do users need to create subscriptions)?

fuyufjh (approved):

PR #89, opened 2024-05-13
RFC: Spill Hash Aggregation

chenzl25:
Tracking issue: https://github.com/risingwavelabs/risingwave/issues/16615


fuyufjh:
The design looks good to me. 👍

I am thinking about another approach. Instead of adapting the HashAgg/HashJoin to splillable by using partitioning, we might just replace the in-memory hash table with some disk-based structure, such as B-Tree e.g. [sled](https://github.com/spacejam/sled), LSM-Tree e.g. RocksDB or on-disk hash table e.g. [odht](https://lib.rs/crates/odht). The major benefit of this disk-based approach is that it **doesn't consume any memory** except some page cache, which is always safe to evict by the OS, so that the memory manager doesn't need to worry about batch memory consumption at all.

I do agree that hybrid hash agg/join (this proposal) is the most commonly used approach in database area, which implies it may have the best performance when memory is not a bottleneck. But I am also curious about the performance gap between it and the disk-based approach. If the performance gap is small, it might be a good fit for RisingWave's case.

chenzl25:
> I am thinking about another approach. Instead of adapting the HashAgg/HashJoin to be spill-able by using partitioning, we might just replace the in-memory hash table with some disk-based structure, such as B-Tree e.g. [sled](https://github.com/spacejam/sled), LSM-Tree e.g. RocksDB or on-disk hash table e.g. [odht](https://lib.rs/crates/odht).

Some drawbacks I come up with using a disk-based structure:
1. We can't monitor the memory consumed by this structure precisely and monitoring the memory consumption in batch query is important, because whether to spill or throw an OOM error to users is determined by the memory monitor.
2. Considering the data size of the current query is too large to be resident in memory, the IO should be the bottleneck. Therefore, using sequential IO or random IO matters. Using a disk-based structure potentially would introduce random IOs which perhaps could be mitigated with OS page cache, but I think when memory is not enough, page cache might not help in this case.
3. Monitoring the disk IO is also important after spilling is enabled. With a disk-based structure, it might be hard to monitor the actual IO.

rfcs/0089-spill-hash-aggregation.md
```diff
@@ -0,0 +1,98 @@
...
+Spill file format:
+
+```
+[proto_len]
+[proto_bytes]
+...
+[proto_len]
+[proto_bytes]
+
+```
```
fuyufjh, line 81: I am not very sure, should we add an CRC here? Personally, I tend to add CRC for any persisted data.

chenzl25: Sounds good, we can add a CRC at the end of the file.

rfcs/0089-spill-hash-aggregation.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Design
+
+As we know a batch query would be fragmented into different fragments and a fragment would be further assigned a parallelism to execute, for each parallelism, we call a task. Hash aggregation is an executor within a task, aka, HashAggExec. To spill a HashAgg, we need to spill the hash table and the input chunks. Ideally, we want HashAggExec to work normally when memory is sufficient and when memory is not enough, we spill the hash table and all input chunks.
+
+![spill-hash-agg](images/0089-spill-hash-aggregation/spill-hash-agg.png)
+
+### Partitions
+
+First, we need to choose a partition number to partition the hash table and input chunks. After partitioning, theoretically, each partition would only contain 1/partition_num of the original data. If this size could be fitted in the memory, we can process the HashAgg partition by partition. If this size is still too large to be fitted in the memory, we need to recursively apply the spill algorithm. When recursively applying the spill algorithm, we need to make sure they use different hash functions to avoid data skew.
```
kwannoel, line 24: >  When recursively applying the spill algorithm, we need to make sure they use different hash functions to avoid data skew.

Could you elaborate on this?

rfcs/0089-spill-hash-aggregation.md
```diff
@@ -0,0 +1,98 @@
...
+  HashAggExec.execute:
+
+    // process hash agg in memory
+    ...
+
+    if not enough memory:
+      // Spill
+      for (k,v) in hash_table:
+        partition = hash(k) % partition_num
+        write_to_agg_state_file(k, v)
```
kwannoel, line 39:
```suggestion
        write_to_agg_state_file(partition, k, v)
```

Is it writing to the partition file?

rfcs/0089-spill-hash-aggregation.md
kwannoel: Why not just use vnode, instead of a separate partition strategy?

chenzl25: > Why would we need to recursively apply the spill algorithm?

Because we don't know how much data the input has. Even if we have 20 partitions by default, a single partition data size still could be too large to fit in memory, so we need to spill the partition again, i.e. recursively apply.

chenzl25: > Why not just use vnode, instead of a separate partition strategy?
For batch query, we don't always have a vnode. Vnode is only associated with the data within a table, but batch hash join input could be intermediate data.

fuyufjh (approved):

PR #91, opened 2024-05-31
RFC: Spill Hash Join

chenzl25:


fuyufjh (approved):

CAJan93:
> Local Disk
Pros: low latency, fixed costs, k8s could handle orphan spilled files in case of failure.

This depends. If RisingWave writes the data to a PersistentVolume (PV) and RW crashes and restarts the pod may bind to the same PV. The data will still be there.

